PR #1, opened 2019-05-22
WAZO-756 conference list

pc-m:
This branch adds the ability to list conference rooms that are configured on an wazo-platform and that are available to a user.

this resource is meant to be used in conjunction with the `/directories/<profile>/sources` which lists the sources that are available to the user.

### DIFF
--- a/wazo_dird/plugins/conference_backend/api.yml
+++ b/wazo_dird/plugins/conference_backend/api.yml
@@ -6,6 +6,7 @@ paths:
       description: '**Required ACL:** dird.backends.conference.sources.read'
       tags:
         - configuration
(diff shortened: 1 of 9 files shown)


wazo_dird/plugins/conference_backend/api.yml
```diff
@@ -110,7 +115,79 @@ paths:
           $ref: '#/responses/ResourceDeleted'
         '404':
           $ref: '#/responses/NotFoundError'
+  /backends/conference/sources/{source_uuid}/contacts:
```
bloom1: I think the CHANGELOG is missing for this route

PR #2, opened 2019-05-27
Wazo 747 default profile for each tenant

pc-m:


fblackburn1:
unittests don't pass

pc-m:
> unittests don't pass

fixed thanks

PR #7, opened 2019-06-17
wazo user contact list: add the id field

pc-m:
The "id" is required to be able to set the contact as a favorite because when
doing a GET /lookup the source_entry_id is mapped to the "id" column

### DIFF
--- a/wazo_dird/plugins/wazo_user_backend/api.yml
+++ b/wazo_dird/plugins/wazo_user_backend/api.yml
@@ -149,6 +149,9 @@ paths:
 definitions:
   WazoContact:
     properties:
+      id:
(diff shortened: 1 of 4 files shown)


wazo_dird/plugins/wazo_user_backend/api.yml
```diff
@@ -149,6 +149,9 @@ paths:
 definitions:
   WazoContact:
     properties:
+      id:
+        type: number
```
bloom1: I think the `number` type is usually used for `double` and `float` types.

pc-m: you're right

PR #9, opened 2019-07-08
WAZO-889-integrate-microsoft-plugin

pc-m:


wazo_dird/plugins/office365_backend/plugin.py
```diff
@@ -0,0 +1,161 @@
...
+
+        for contact in updated_contacts:
+            if self._first_match_predicate(lowered_term, contact):
+                return self._SourceResult(contact)
+
+    def _first_match_predicate(self, term, contact):
+        for column in self._first_matched_columns:
+            column_value = contact.get(column) or ''
+
+            if not isinstance(column_value, list):
```
sileht: Starting with the True expression first is more intuitive, e.g:

`if isinstance(column_value, list):`

wazo_dird/plugins/office365_backend/schemas.py
```diff
@@ -0,0 +1,35 @@
...
+
+class SourceSchema(BaseSourceSchema):
+
+    auth = fields.Dict(
+        keys=fields.String(validate=Length(min=1, max=64)),
+        values=fields.String(validate=Length(min=1, max=64))
+    )
+    endpoint = fields.String(
+        missing='https://graph.microsoft.com/v1.0/me/contacts',
+        validate=Length(min=1, max=255),
```
sileht: Does Microsoft provides this service on premise or we have something only for testing in our API ?

wazo_dird/plugins/office365_backend/services.py
```diff
@@ -0,0 +1,75 @@
...
+            "search": term
+        }
+        try:
+            response = requests.get(url, headers=headers, params=query_params)
+            if response.status_code == 200:
+                logger.debug('Sucessfully fetched contacts from microsoft.')
+                return response.json().get('value', [])
+            else:
+                return []
+        except requests.exceptions.RequestException as e:
```
sileht: `requests.RequestException` is enough if you want to save some char.

wazo_dird/plugins/office365_backend/plugin.py
pc-m: I agree

wazo_dird/plugins/office365_backend/schemas.py
pc-m: I don't know if Microsoft offers this service on premise. I think the parameter was there when the proof of concept was written, then it got into the tests and made it to production.

I would not mind replacing the argument with a config option to make the API simpler in the future.

pc-m:
> Oh I just see that debian new dependencies are missing.

I add python3-marshmallow

pc-m:
> requirements.txt need to be updated too.

I cannot find which ones are missing for dird, I will check wazo-auth

PR #13, opened 2019-08-20
WAZO-1001 Add default_json lookup by user UUID

bloom1:


fredericlepied:
recheck

wazo_dird/plugins/default_json/api.yml
```diff
@@ -29,6 +29,37 @@ paths:
...
+
+        The `lookup` query will return a list of result matching the searched term.
+        The result will be retrieved from all configured directories for the given
+        profile and user.
+
+
+
+        This route is provided by the `default_json_view` plugin using the `lookup`
+        plugin and all configured sources for the given profile.'
+      operationId: lookup
```
pc-m: I think the `operationId` needs to be different for all resources in the spec

wazo_dird/plugins/default_json/http.py
```diff
@@ -77,6 +78,57 @@ def get(self, profile):
         return response
 
 
+class LookupByUUID(LegacyAuthResource, DisplayAwareResource):
```
pc-m: You should use `AuthResource` instead of `LegacyAuthResource`
The only difference is that `AuthResource` will return errors in the "new" format and legacy is used for the old API to avoid bumping the version

wazo_dird/plugins/default_json/http.py
```diff
@@ -77,6 +78,57 @@ def get(self, profile):
...
+    def __init__(
+        self, lookup_service, favorite_service, display_service, profile_service, auth_client
+    ):
+        self.lookup_service = lookup_service
+        self.favorite_service = favorite_service
+        self.display_service = display_service
+        self.profile_service = profile_service
+        self.auth_client = auth_client
+
+    @required_acl('dird.directories.lookup.{profile}.{xivo_user_uuid}.read')
```
pc-m: remove the `user_uuid` prefix to avoid adding a xivo in our public API

wazo_dird/plugins/default_json/http.py
```diff
@@ -77,6 +78,57 @@ def get(self, profile):
...
+    def get(self, profile, xivo_user_uuid):
+        args = parser.parse_args()
+        term = args['term']
+
+        logger.info('Lookup %s for user %s with profile %s', term, xivo_user_uuid, profile)
+
+        try:
+            tenant_uuid = self._get_user_tenant_uuid(xivo_user_uuid)
```
pc-m: Using the tenant_uuid based on the user_uuid parameter allow a malicious user with the right ACL to do queries using the user_uuid of users in another tenant.

You should use the token's tenant_uuid or check that the user doing the query is above or in the user (from the parameter) tenant_uuid

wazo_dird/plugins/default_json/http.py
```diff
@@ -77,6 +78,57 @@ def get(self, profile):
...
+            logger.debug('Got tenant_uuid %s for user %s, profile %s', tenant_uuid, xivo_user_uuid, profile)
+            profile_config = self.profile_service.get_by_name(tenant_uuid, profile)
+            display = self.build_display(profile_config)
+        except OldAPIException as e:
```
pc-m: I did not check but the errors should have the new style of body

wazo_dird/plugins/default_json/api.yml
pc-m: From the swagger editor
```
Semantic error at paths./directories/lookup/{profile}/{xivo_user_uuid}.get.operationId
Operations must have unique operationIds.
```
Unfortunately there are so many errors that its hard to see when we introduce new errors

wazo_dird/plugins/default_json/http.py
bloom1: Do I do the same in the swagger doc?

pc-m: nevermind, we did not create another key without the prefix in the token data :(

bloom1: Ah that's why I tried and it kept failing on me!

bloom1:
recheck

bloom1:
recheck

PR #14, opened 2019-08-21
tests: fix sessionmaker binding

sileht:


pc-m:
recheck

sileht:
Looks like there is other place where Metadata is expected to be binded because some sql request are done outside a session. Since I don't want to dig deeper now, and just want the buster branch to work. I just fix the session_maker() binding.

sileht:
recheck

PR #20, opened 2019-09-19
dird does not need to start after confgend

pc-m:
dird used to call xivo-confgen to get its configuration but not anymore


pc-m:
> We should also delete dird section in wazo-confgend to avoid misunderstanding

I'll look at that this sprint, I kept it around because I might use it to migrate some of the old sources to the new ones, ldap sources for example.

PR #37, opened 2019-11-22
backend wazo contacts: allow to pass many uuid to get users

fblackburn1:
Depends-on: https://github.com/wazo-platform/wazo-confd/pull/43
Depends-on: https://github.com/wazo-platform/wazo-dird-client/pull/9

### DIFF
--- a/wazo_dird/plugins/wazo_user_backend/api.yml
+++ b/wazo_dird/plugins/wazo_user_backend/api.yml
@@ -126,7 +126,7 @@ paths:
         - $ref: '#/parameters/limit'
         - $ref: '#/parameters/offset'
         - $ref: '#/parameters/search'
-        - $ref: '#/parameters/backend_wazo_uuid'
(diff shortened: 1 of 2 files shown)


wazo_dird/plugins/wazo_user_backend/api.yml
```diff
@@ -126,6 +126,7 @@ paths:
         - $ref: '#/parameters/limit'
         - $ref: '#/parameters/offset'
         - $ref: '#/parameters/search'
+        - $ref: '#/parameters/backend_wazo_uuid'
```
bloom1: I think it would be better to call it `wazo_backend_uuid` instead (even though it's only useful in the spec). `backend_wazo_uuid` gives the impression that it is about the Wazo UUID.

fblackburn1: fixed

bloom1:
recheck

PR #41, opened 2019-12-10
WAZO-1081: Google more contact info

bloom1:
Extract and expose more info from the Google source:

- [x] First Name
- [x] Name
- [x] Enterprise
- [x] Function
- [x] All tel and label
- [x] All Email and label
- [x] All addresses
- [x] Note

### DIFF
--- a/integration_tests/suite/test_google_contacts.py
+++ b/integration_tests/suite/test_google_contacts.py
@@ -57,6 +57,16 @@ GOOGLE_CONTACT_LIST = {
                         "$t": "+1 5555551111",
                     },
                 ],
+                "gd$structuredPostalAddress": [
+                    {
+                        "rel": "http://schemas.google.com/g/2005#home",
+                        "gd$formattedAddress": {"$t": "Main Land"},
+                    },
+                    {
+                        "label": "Second address",
+                        "gd$formattedAddress": {"$t": "Alternative Land"},
+                    },
+                ],
             },
             {
                 "id": {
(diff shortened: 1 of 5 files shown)


integration_tests/suite/test_google_contacts.py
```diff
@@ -160,17 +183,25 @@ def test_list(self, google_api):
...
                         emails=contains_inanyorder('mario@bros.example.com'),
+                        email_labels=contains_inanyorder('other'),
                         numbers=contains_inanyorder('+15555551111', '+15555551234'),
                         numbers_by_label=has_entries(
                             home='+15555551111', mobile='+15555551234'
                         ),
+                        organizations=contains_inanyorder('MarioLand'),
+                        organization_titles=contains_inanyorder('Artist'),
+                        addresses=contains_inanyorder('Main Land', 'Alternative Land'),
+                        address_labels=contains_inanyorder('home', 'Second address'),
```
fblackburn1: After IRL discussion, labels should be merged with objects inside a list of dictionnary, not a list of string

bloom1:
recheck

fblackburn1:
recheck

bloom1:
recheck

bloom1:
recheck

bloom1:
@fblackburn1 it failed because of the display that uses the old email format - would it be better to leave the old emails field and add emails_with_labels or something like that? Or to update the display?

wazo_dird/plugins/google_backend/plugin.py
```diff
@@ -121,9 +121,15 @@ def _first_match_predicate(self, term, contact):
...
-                        return True
+                    if isinstance(value, dict):
+                        for sub_value in value.values():
+                            if term == sub_value.lower():
+                                return True
+                    else:
+                        if term == value.lower():
+                            return True
             else:
+                logger.debug('Item is not a dict or list')
```
fblackburn1: We can remove it.
But just wait until zuul complete its check and we will force push/merge if need to be in this release

bloom1: Oops, did not mean to leave it in. I will force push without it.

PR #44, opened 2020-01-16
integration tests: fix race condition for late_confd

sduthil:
Why:

* Sometimes, the test would fail because late_confd would start too
early in the test


integration_tests/assets/confd_data/asset.wazo_users_late_confd/run_confd
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 
-sleep 10
```
pc-m: I LOVE seeing a `sleep 10` disappear. Thanks!

PR #49, opened 2020-02-27
WAZO-1375 csv backend: log errors when an absent column is configured

pc-m:
If a column is configured in searched_columns, unique_column,
first_match_columns and that column does not exist dird will ignore it during a
lookup and log an error message.

### DIFF
--- a/wazo_dird/plugins/source_result.py
+++ b/wazo_dird/plugins/source_result.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2014-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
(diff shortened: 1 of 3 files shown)


pc-m:
recheck

wazo_dird/plugins/source_result.py
```diff
@@ -64,9 +64,10 @@ def get_unique(self):
         try:
             return str(self.fields[self._unique_column])
         except KeyError:
-            logger.exception(
-                '%s is not properly configured, the unique column is not part of the result',
+            logger.info(
```
sduthil: Why `logger.info`? How about `logger.error`?

pc-m:
@sduthil are there anything missing in this PR to include in the current sprint?

sduthil:
> @sduthil are there anything missing in this PR to include in the current sprint?

Nothing missing, I just wanted some time to play with it, but could not find the time.

PR #50, opened 2020-03-06
WAZO-1483: fix office365 search

bloom1:
alembic: add office365 default values
office365: fix OpenAPI spec and add description
office365: fix unhandled exception
config_service: fix office365 default config
office365_backend: aggregate all contact numbers in a field
office365_backend: remove search with term

### DIFF
--- a/wazo_dird/plugins/office365_backend/plugin.py
+++ b/wazo_dird/plugins/office365_backend/plugin.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2019-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
@@ -78,9 +78,7 @@ class Office365Plugin(BaseSourcePlugin):
         except MicrosoftTokenNotFoundException:
             return []
 
-        contacts = self.office365.get_contacts_with_term(
-            microsoft_token, term, self.endpoint
-        )
+        contacts = self.office365.get_contacts(microsoft_token, self.endpoint)
         updated_contacts = self._update_contact_fields(contacts)
 
         lowered_term = term.lower()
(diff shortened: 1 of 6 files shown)


wazo_dird/plugins/office365_backend/plugin.py
```diff
@@ -77,8 +77,11 @@ def search(self, term, args=None):
             microsoft_token = self._get_microsoft_token(**args)
         except MicrosoftTokenNotFoundException:
             return []
+        try:
+            contacts = self.office365.get_contacts(microsoft_token, self.endpoint)
+        except UnexpectedEndpointException:
```
pc-m: I think something should be logged here to help debugging when there's no result

wazo_dird/plugins/office365_backend/plugin.py
```diff
@@ -101,7 +104,11 @@ def list(self, unique_ids, args=None):
         except MicrosoftTokenNotFoundException:
             return []
 
-        contacts = self.office365.get_contacts(microsoft_token, self.endpoint)
+        try:
+            contacts = self.office365.get_contacts(microsoft_token, self.endpoint)
+        except UnexpectedEndpointException:
+            return []
```
pc-m: Some logging here too

wazo_dird/plugins/office365_backend/plugin.py
```diff
@@ -125,7 +132,11 @@ def first_match(self, term, args=None):
             )
             return None
 
-        contacts = self.office365.get_contacts(microsoft_token, self.endpoint)
+        try:
+            contacts = self.office365.get_contacts(microsoft_token, self.endpoint)
+        except UnexpectedEndpointException:
+            return None
```
pc-m: logging again

wazo_dird/plugins/office365_backend/plugin.py
bloom1: There is already a logger.error in the get_contacts method before the exception raising. Maybe we should only include more details? All it says is `An error occured while fetching information from microsoft endpoint`

pc-m: You're right I missed that. We could add the status code to the current log line and maybe log the response body in a logger.debug?

pc-m: I'll merge anyway since that's not something that changed from your PR.

PR #55, opened 2020-04-16
reverse: dont cancel other reverse upon failure

sduthil:
Why:

* If one reverse source fails, it would cancel other reverse sources,
because `[]` is not `None`.

How:

* The tests try to reverse lookup on a broken source first, then on a
valid source. The code does not guarantee 100% that the broken source runs
first, but the lock coupled with time.sleep should be enough.


fblackburn1:
recheck

PR #61, opened 2020-05-01
Pull request for WAZO-1747-confd-http

fblackburn1:
update directories config to use confd without ssl
test: fix python segfault

### DIFF
--- a/alembic/versions/0542e068b5ef_migrate_confd_config_without_ssl.py
+++ b/alembic/versions/0542e068b5ef_migrate_confd_config_without_ssl.py
@@ -38,6 +38,7 @@ def upgrade():
             source.extra_fields['confd'].pop('verify_certificate', None)
         else:
             source.extra_fields['confd']['prefix'] = '/api/confd'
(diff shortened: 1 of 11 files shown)


alembic/versions/0542e068b5ef_migrate_confd_config_without_ssl.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if (
+            confd_config['host'] in ('localhost', '127.0.0.1')
+            and confd_config['port'] == 9486
+        ):
+            source.extra_fields['confd']['prefix'] = None
+            source.extra_fields['confd']['https'] = False
+            source.extra_fields['confd'].pop('verify_certificate', None)
+        else:
+            source.extra_fields['confd']['prefix'] = '/api/confd'
```
pc-m, line 40: If the host is `myhost.example.com` and port `9486`
You should add the prefix AND remove the port.

fblackburn1: fixed

PR #62, opened 2020-05-05
Pull request for WAZO-1720-graphql-reverse-lookup

sduthil:
Allow reverse lookup operation via GraphQL HTTP endpoint

Depends-On: https://github.com/wazo-platform/wazo-dird-client/pull/12


sduthil:
recheck

PR #70, opened 2020-05-21
improve docker size

fblackburn1:


Dockerfile
```diff
@@ -1,25 +1,30 @@
...
+RUN apt-get -q update
+RUN apt-get -yq install gcc libldap2-dev libsasl2-dev
 
+COPY . /usr/src/wazo-dird
 WORKDIR /usr/src/wazo-dird
+RUN pip install -r requirements.txt
+RUN python setup.py install
+
+FROM python:3.7-slim-buster AS build-image
+COPY --from=compile-image /opt/venv /opt/venv
```
bloom1, line 17: I added 
```
RUN apt-get -q update
RUN apt-get -yq install libldap-2.4-2
```
here to fix the missing ldap shared library issue, but I'm not sure if this is the right place to do it. Can you confirm that this would be the correct solution?

Thank you!

fblackburn1: Installing libldap seems the correct solution but not at the right place.
Each RUN/cCOPY/ADD will create a new layer (and increase size). We should add it to the existing RUN below

bloom1:
recheck

PR #76, opened 2020-07-20
wazo backend: add default reverse lookup on "exten" field

sduthil:
Why:

* An application may want to know who is the caller from call logs.


fblackburn1:
regate

PR #91, opened 2020-12-11
Pull request for WAZO-2149-implement-match-all-for-ldap

fblackburn1:


wazo_dird/plugins/ldap_backend/plugin.py
```diff
@@ -52,6 +52,22 @@ def first_match(self, term, args=None):
 
         return self._first_match_and_format(filter_str)
 
+    def match_all(self, extens, args=None):
+        results = {}
+        columns = self._ldap_config.first_matched_columns()
+        logger.debug(f'Looking for columns ({columns}) with ({extens})')
```
pc-m, line 58: You should not use string formatting in loggers.

If you not in debug mode the string will substitute `columns` and `extens` event if it's not going to be logged

fblackburn1: yep ... that's black effect, I was focus to fit my logger in one line :confounded: 
And your explanation it's the github review effect ... you know I know it, but you need to explain why to be 100% sure :confounded:

pc-m (approved):

PR #93, opened 2021-01-07
Pull request for WAZO-2177-duplicate-mobile

sduthil:
## office365: add numbers_except_label

Why:

* avoid duplication mobile phone numbers

## office365: split phone fields with single/multiple values

### DIFF
--- a/wazo_dird/plugins/google_backend/services.py
+++ b/wazo_dird/plugins/google_backend/services.py
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2019-2021 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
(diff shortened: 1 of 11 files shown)


wazo_dird/plugins/google_backend/services.py
```diff
@@ -185,6 +186,28 @@ def _extract_numbers_by_label(cls, contact):
...
+            'home',
+            'work',
+            'mobile',
+            'other',
+            'main',
+            'home_fax',
+            'work_fax',
+            'google_voice',
+            'pager',
+        ):
```
fblackburn1, line 204: Should be possible to use keys?
```suggestion
        for type_ in numbers_by_label.keys():
```

sduthil: No: in case the contact has only a `home` phone, the `mobile` key will not be present in the google contact, and will not be present either in `numbers_by_label`. So `numbers_except_label[mobile][0]` will return nothing (due to a `KeyError: mobile`) instead of returning the `home` phone

fblackburn1 (approved):

PR #96, opened 2021-02-17
upgrade office356: accept missing default format column

sduthil:
Why:

* Office365 sources may not have a "phone" format column


pc-m (commented):
Assigning a string to `None`? Is that possible?

pc-m:
> Assigning a string to `None`? Is that possible?

Sorry I missread

pc-m (approved):

PR #102, opened 2021-04-15
graphql: avoid failing when a column is missing from a source

pc-m:


fblackburn1 (commented):
I'm able to make the test fail if I revert the change
Have you try to make it fail?

I can see the traceback disappear, but the added test doesn't fail

pc-m:
> I'm able to make the test fail if I revert the change
> Have you try to make it fail?
> 
> I can see the traceback disappear, but the added test doesn't fail

You're right, I did not see the the result was in the "data" of the result when the error occurred :(
It will make the log file less noisy but does not really fix anything unfortunately

fblackburn1 (approved):
Thank you :+1:

PR #103, opened 2021-04-15
office/google: do not use logger.error in normal situations

pc-m:
when a user has not logged into office or google
when an administrator has not configured office or google applications

### DIFF
--- a/wazo_dird/plugins/google_backend/services.py
+++ b/wazo_dird/plugins/google_backend/services.py
@@ -98,6 +98,16 @@ def get_google_access_token(user_uuid, wazo_token, **auth_config):
         auth = Auth(token=wazo_token, **auth_config)
         return auth.external.get('google', user_uuid).get('access_token')
     except requests.HTTPError as e:
+        if e.response.status_code == 404:
+            if 'unknown-external-auth-type' in e.response.text:
+                logger.debug('The "google" authentication type has not been configured')
+                raise GoogleTokenNotFoundException(user_uuid)
+            elif 'unknown-external-auth' in e.response.text:
+                logger.debug(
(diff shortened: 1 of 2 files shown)


wazo_dird/plugins/google_backend/services.py
```diff
@@ -98,6 +98,16 @@ def get_google_access_token(user_uuid, wazo_token, **auth_config):
...
+                logger.debug('The "google" authentication type has not been configured')
+                raise GoogleTokenNotFoundException(user_uuid)
+            elif 'unknown-external-auth' in e.response.text:
+                logger.debug(
+                    'user %s has no "google" authentication configured', user_uuid
+                )
+                raise GoogleTokenNotFoundException(user_uuid)
+
         logger.error('Google token could not be fetched from wazo-auth, error: %s', e)
         raise GoogleTokenNotFoundException(user_uuid)
```
fblackburn1, line 112:
```suggestion
        if e.response.status_code == 404:
            if 'unknown-external-auth-type' in e.response.text:
                logger.debug('The "google" authentication type has not been configured')
            elif 'unknown-external-auth' in e.response.text:
                logger.debug(
                    'user %s has no "google" authentication configured', user_uuid
                )
            else:
                logger.error('Google token could not be fetched from wazo-auth, error: %s', e)

        raise GoogleTokenNotFoundException(user_uuid)
```

pc-m: I also want the error logged if it's not a 404

fblackburn1: Oh didn't see, you're right

fblackburn1 (approved):

PR #109, opened 2021-07-07
WAZO-2370 Contacts->People API

benbdevd:


benbdevd:
Branch name writes stdout to a file 'People API', bad naming, don't do this

PR #110, opened 2021-07-08
Wazo 2370 people api migration

benbdevd:


benbdevd:
Old group fetching is no longer needed because of changes to the API.

It is no longer necessary to query for group ID because the function would only search for the 'My Contacts' system group, which has now been assigned to the unchanging `contactGroups/myContacts`
https://developers.google.com/contacts/v3/reference#gcSystemGroup

It is not necessary to request the contacts in the 'My Contacts' system group because it will always be a subset of the default listing of all 'connections' with `people.connections.list`

pc-m (changes requested):
It works great, nice work. I've added some inline comments for things that could be improved. I also think you should add an upgrade note here https://wazo-platform.org/uc-doc/upgrade/upgrade_notes for the change in permissions that is required for the plugin to keep working after the upgrade.
I also THINK that users will have to re-enable the google plugin once the permissions have been changed to accept the new requirements of the plugin, this should also be in the changelog if you can confirm that this is what happens.

wazo_dird/plugins/google_backend/services.py
```diff
@@ -36,39 +42,58 @@ def get_contacts(self, google_token, **list_params):
...
+            self.batch_url, headers, params, 'Fetched batch of contacts from Google'
         )
+        return response.json().get('responses', [])
+
+    def _get_request(self, url, headers, params, debug_message=None):
+        # Requests have verify=False because the integration test mock servers do not have proper SSL
+        # Find a way to selectively turn off verification during testing only
+        response = requests.get(url, headers=headers, params=params, verify=False)
         if response.status_code != 200:
+            logger.debug('Get Request Unsuccessful: %s' % str(response.status_code))
```
pc-m, line 90: You should not use `%` to format strings. The `%s` also does the `str()` on the parameter. This should be
```python
logger.debug('Get Request Unsuccessful: %s', response.status_code)
```
Same thing for the other logger.debug.

You can safely forget about the `%` operator for string formating ;)

pc-m (changes requested):

wazo_dird/plugins/google_backend/services.py
```diff
@@ -218,37 +242,39 @@ def _extract_numbers_except_label(cls, contact):
...
 
     @classmethod
     def _extract_type(cls, entry):
-        rel = entry.get('rel')
-        if rel:
-            _, type_ = rel.rsplit('#', 1)
-        else:
-            type_ = entry.get('label')
-        return type_
+        return entry.get('type', 'other')
```
wavecommander, line 269: Not sure if this should stay like this, or move the 'other' into `extract_numbers_by_label()`

wavecommander: This does seem to fix [WAZO-2439](https://wazo-dev.atlassian.net/browse/WAZO-2439)

pc-m (approved):

PR #116, opened 2021-12-15
Wazo-2633: Automatic reloading of csv files in case it has been updated by client

nissou91:


wazo_dird/plugins/csv_backend/plugin.py
```diff
@@ -105,6 +111,16 @@ def _load_file(self):
         except IOError:
             logger.exception('Could not load CSV file content')
 
+    def _check_for_csv_file_update(self, filename):
+        fname = pathlib.Path(filename)
+        csv_file_last_modification_date = datetime.datetime.fromtimestamp(
+            fname.stat().st_mtime
+        )
+        current_date = datetime.datetime.today()
```
pc-m, line 119: I don't think you need to do any time arithmetic. If `csv_file_last_modification_date` is different from the last modification time from the last time the file got loaded you can re-read the file and save the modification time for your next check.

pc-m (changes requested):
You need to implement a test for this new scenario in the integration tests here https://github.com/wazo-platform/wazo-dird/blob/master/integration_tests/suite/test_csv_backend.py
In you test you can use the ` self.docker_exec` command to execute a command in the running container to modify the CSV file.

wazo_dird/plugins/csv_backend/plugin.py
```diff
@@ -34,6 +35,10 @@ class CSVPlugin(BaseSourcePlugin):
     The `searched_columns` are the columns used to search for a term
     """
 
+    def __init__(self) -> None:
```
pc-m, line 38: We do not use type annotations in Wazo at this time. Not that I have anything against it but it's a change that should be discussed before using them.
```suggestion
    def __init__(self):
```

pc-m: Alternatively, the `_csv_last_modification_time` initialisation could also be moved to the `load` function

wazo_dird/plugins/csv_backend/plugin.py
```diff
@@ -105,6 +114,14 @@ def _load_file(self):
         except IOError:
             logger.exception('Could not load CSV file content')
 
+    def _check_for_csv_file_update(self, filename):
+        fname = pathlib.Path(filename)
+        tmp_csv_file_last_modification_date = fname.stat().st_mtime
+
+        if self._csv_last_modification_time != tmp_csv_file_last_modification_date:
+            self._load_file()
+            self._csv_last_modification_time = tmp_csv_file_last_modification_date
```
pc-m, line 123: During the `load` of the plugin `_load_file` is called (line 50) and the `_csv_last_modification_time` is not updated. I think this line should be moved to the `_load_file` function. And maybe move all of this logic to `_load_file` and call it from each functions.

integration_tests/suite/test_csv_backend.py
```diff
@@ -39,6 +40,38 @@ def test_that_searching_for_lice_return_Alice(self):
             contains(has_entries(column_values=self._alice + favorite)),
         )
 
+    def test_that_csv_file_changes_have_been_committed(self):
```
pc-m, line 43: Very nice test, exactly what was needed.

pc-m (changes requested):

wazo_dird/plugins/csv_backend/plugin.py
```diff
@@ -94,16 +100,27 @@ def _load_file(self):
...
-            logger.debug('Reading %s with delimiter %r', filename, delimiter)
-            with open(filename, 'r') as f:
-                csvreader = csv.reader(f, delimiter=delimiter)
-                keys = [key for key in next(csvreader)]
-                self._content = [self._row_to_dict(keys, row) for row in csvreader]
-                logger.debug('Loaded with %s', self._content)
-        except IOError:
-            logger.exception('Could not load CSV file content')
+            tmp_csv_file_last_modification_date = fname.stat().st_mtime
+            if self._csv_last_modification_time != tmp_csv_file_last_modification_date:
```
pc-m, line 106: Could you change the condition to return early and reduce indentation
```python
if self._csv_last_modification_time == tmp_csv_file_last_modification_date:
    return
```

wazo_dird/plugins/csv_backend/plugin.py
```diff
@@ -94,16 +100,27 @@ def _load_file(self):
...
+                        csvreader = csv.reader(f, delimiter=delimiter)
+                        keys = [key for key in next(csvreader)]
+                        self._content = [
+                            self._row_to_dict(keys, row) for row in csvreader
+                        ]
+                        logger.debug('Loaded with %s', self._content)
+                except IOError:
+                    logger.exception('Could not load CSV file content')
+                finally:
+                    self._csv_last_modification_time = (
```
pc-m, line 119: I think this line needs to be in the `try` block. If an `IOError` we do not want to save the last modification time since we failed to read the file.

pc-m (changes requested):
Nice job Aniss, can you do a `git rebase -i master` and clean up your commits. Some of them have no commit message and other could be squashed together.

PR #127, opened 2022-02-21
pin markupsafe version to buster

fblackburn1:
why: if not pinned, pip install latest version (2.1.0) which is
incompatible with flask 1.0.2 from buster


fblackburn1:
recheck

PR #131, opened 2022-03-28
Update is_valid function call with new name

nissou91:


pc-m (approved):

pc-m (changes requested):
I think this one should use `is_valid`

PR #149, opened 2022-12-06
Pull request for zuul-debian-packaging

fblackburn1:
## zuul: validate debian packaging


fblackburn1:
debian-packaging was sucess
But don't want to wait 2 hours for integration tests

PR #157, opened 2023-04-05
Add indices to Foreign Keys

sopelj:


fblackburn1 (dismissed):

sopelj:
> recheck (to validate it's not a CI issue)

It's not just Zuul, they're failing locally too. Although, I have no clue why. I'm looking into it.

PR #165, opened 2023-06-02
Wp 926 sort personal contacts

xxbillouxx:


xxbillouxx:
recheck

DrPyser:
Checked that API output seems ordered appropriately in e2e test :+1: 
Unfortunately many integration tests are flaky on my end.

DrPyser:
> Checked that API output seems ordered appropriately in e2e test +1 Unfortunately many integration tests are flaky on my end.

Okay, was missing a `git pull`. Tests are passing fine.

wazo_dird/database/queries/base.py
```diff
@@ -87,3 +101,52 @@ def _get_dird_user(self, session, user_uuid):
...
+                )
+            except KeyError:
+                raise ValueError(f"order: column '{order}' was not found")
+        elif reverse:
+            rows.reverse()
+
+        if not limit:
+            return rows[offset:]
+        else:
+            return rows[offset : offset + limit]
```
DrPyser, line 129: since we are returning a slice anyway, I would avoid mutating the `rows` parameter(or mutating parameters in general), which might cause issues if the list passed to this method is otherwise reused.
https://en.wikipedia.org/wiki/Referential_transparency

instead of `rows.reverse`, `rows = reversed(rows)` would produces an iterator, and `itertools.islice` would slice any iterable or iterator and should minimize copying.

```
if order:
    rows = ...
elif reversed:
    rows = reversed(rows)

return list(itertools.islice(rows, offset, offset + limit if limit else None))
```
here reversed does no copying, and a new list is generated by the `list` call.

wazo_dird/database/queries/base.py
```diff
@@ -87,3 +101,52 @@ def _get_dird_user(self, session, user_uuid):
...
+
+        if parameters['limit'] is not None and int(parameters['limit']) <= 0:
+            raise ValueError('limit must be a positive number')
+
+        if parameters['direction'] not in self.SORT_DIRECTIONS.keys():
+            raise ValueError('direction must be asc or desc')
+
+    def _extract_search_params(self, parameters={}):
+        parameters = parameters.copy()
+        parameters = self._populate_parameters(parameters)
```
DrPyser, line 143: you're copying explicitly, which is good for referential transparency, but `self._populate_parameters` also creates a new dict anyway, so there some superfluous copying here.

if you assume(and guarantee) that your methods avoid mutating their parameters, you don't need to explicitly copy.

wazo_dird/database/queries/base.py
```diff
@@ -87,3 +101,52 @@ def _get_dird_user(self, session, user_uuid):
...
+        self.validate_parameters(parameters)
+
+        order = parameters.pop('order', None)
+        limit = parameters.pop('limit', None)
+        if limit is not None:
+            limit = int(limit)
+        offset = int(parameters.pop('offset', 0))
+        reverse = False if parameters.pop('direction', 'asc') == 'asc' else True
```
DrPyser, line 151:
```suggestion
        reverse = not (parameters.pop('direction', 'asc') == 'asc')
```

wazo_dird/database/queries/personal.py
```diff
@@ -86,15 +86,19 @@ def _new_user_contacts_filter(self, user_uuid):
 
 
 class PersonalContactCRUD(BaseDAO):
-    def list_personal_contacts(self, user_uuid=None):
+    def list_personal_contacts(self, user_uuid=None, search_params={}):
```
DrPyser, line 89: one should avoid mutable default parameter values, as the literal `{}` is only evaluated once on import, and the same reference to the created `dict` object gets reused for any subsequent call to the method(hence any mutation on the value will affect all subsequent call with that same default value).
https://docs.python-guide.org/writing/gotchas/
```suggestion
    def list_personal_contacts(self, user_uuid=None, search_params=None):
        search_params = {} if search_params is None else search_params
```

wazo_dird/database/queries/base.py
```diff
@@ -87,3 +101,52 @@ def _get_dird_user(self, session, user_uuid):
...
+        if int(parameters['offset']) < 0:
+            raise ValueError('offset must be positive number')
+
+        if parameters['limit'] is not None and int(parameters['limit']) <= 0:
+            raise ValueError('limit must be a positive number')
+
+        if parameters['direction'] not in self.SORT_DIRECTIONS.keys():
+            raise ValueError('direction must be asc or desc')
+
+    def _extract_search_params(self, parameters={}):
```
DrPyser, line 141: one should avoid mutable default parameter values, as the literal `{}` is only evaluated once on import, and the same reference to the created `dict` object gets reused for any subsequent call to the method(hence any mutation on the value will affect all subsequent call with that same default value).
https://docs.python-guide.org/writing/gotchas/
```suggestion
    def _extract_search_params(self, parameters=None):
        parameters = {} if parameters is None else parameters
```

DrPyser (changes requested):

DrPyser (approved):
:+1: GG

PR #166, opened 2023-06-06
Wp 878 simplification optimization

xxbillouxx:


DrPyser (changes requested):

DrPyser:
The code was improved, but is there evidence that this results in better performance behavior? Is there a benchmark test, if not automated than a straightforward procedure for reproduction?

wazo_dird/plugins/office365_backend/services.py
```diff
@@ -24,105 +24,57 @@ class Office365Service(SelfSortingServiceMixin):
...
                 )
                 raise UnexpectedEndpointException(
                     endpoint=url, error_code=response.status_code
                 )
         except requests.RequestException:
+            logger.error(f'An error occured while fetching data from Microsoft {url}')
             raise UnexpectedEndpointException(endpoint=url)
 
+    def _extract_contacts(self, response):
+        return response.json().get('value', [])
```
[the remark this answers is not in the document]
xxbillouxx: I would prefer to keep this function, to avoid duplicated code

xxbillouxx:
> The code was improved, but is there evidence that this results in better performance behavior? Is there a benchmark test, if not automated than a straightforward procedure for reproduction?

I manualy tested this PR. There is no easy way to test (e2e) the improvement, because the test (to validate) involve the frontend too.

DrPyser:
Tested contact search and first page rendering for a office 365 account with 5000 contacts.
Takes about 9-10 seconds for searching a specific contact and rendering the first page of results.

DrPyser:
Testing with 5000 contacts in outlook account.
Without this PR, dird takes 203.19s to produce a response for the first 30 contacts(the user never sees that response as the gateway timesout before that of course), as the backend seems to be fetching all contacts by groups of 10(?!?).

DrPyser:
With this PR, I see the API response for the first 30 takes 7.60s. Much better.

wazo_dird/plugins/office365_backend/services.py
```diff
@@ -24,105 +24,58 @@ class Office365Service(SelfSortingServiceMixin):
...
 
     def get_contacts(self, microsoft_token, url, **list_params):
-        count, has_multiple_pages = self._get_total_contacts(microsoft_token, url)
-        contacts = list(self._fetch(microsoft_token, url, count, has_multiple_pages))
+        count = self._get_total_contacts(microsoft_token, url)
+        contacts = list(self._fetch(microsoft_token, url, count))
         total_contacts = len(contacts)
         sorted_contacts = self.sort(contacts, **list_params)
         paginated_contacts = self._paginate(sorted_contacts, **list_params)
         return paginated_contacts, total_contacts
```
DrPyser, line 32: The `list_params` could be used for the microsoft API query in order to avoid over-querying for better performance. For example, listing the first 30 contacts would translate to `self._fetch(microsoft_token, url, list_params['limit'])`. The microsoft API seems to support sorting and filtering based on the contact attributes: https://learn.microsoft.com/en-us/previous-versions/office/office-365-api/api/version-2.0/complex-types-for-mail-contacts-calendar#OdataQueryParams.

DrPyser:
Looking up a user takes between 7 and 9 seconds. Also much better.

DrPyser (approved):
There is a significant improvement through this refactoring, so I'll approve it.
Further attempts at improving performance could better exploit the microsoft API to avoid overfetching, or caching could be used to avoid fetching contacts from microsoft for any query(e.g. once contacts are fetched and listed, they should not need to be refetched for every lookup)

DrPyser (approved):
There is a significant improvement through this refactoring, so I'll approve it.
Further attempts at improving performance could better exploit the microsoft API to avoid overfetching, or caching could be used to avoid fetching contacts from microsoft for any query(e.g. once contacts are fetched and listed, they should not need to be refetched for every lookup)

wazo_dird/plugins/office365_backend/services.py
xxbillouxx: Thank you for your feedback.

Be careful, you pointed to the deprecated Outlook API.

There is a quota for the Microsoft Graph API, based on the number of requests. This is why I request all the contacts. Less requests => quota OK.

I'm not sure requesting 1K or 5K contacts from Microsoft will change their request time.

If we use `list_params['limit']` we must "store" somewhere the `@odata.nextLink` (there is no more `skip` parameter)
https://learn.microsoft.com/en-us/graph/paging?context=graph%2Fapi%2F1.0&view=graph-rest-1.0
https://learn.microsoft.com/en-us/graph/api/user-list-contacts?view=graph-rest-1.0&tabs=http

PR #170, opened 2023-07-26
Wazo 3267 phonebook source schema

DrPyser:
https://wazo-dev.atlassian.net/browse/WAZO-3267
depends-on: #167

### DIFF
--- a/integration_tests/suite/test_phonebook_http.py
+++ b/integration_tests/suite/test_phonebook_http.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from contextlib import contextmanager
+import logging
+from uuid import uuid4
 from hamcrest import (
     assert_that,
     calling,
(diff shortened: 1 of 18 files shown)


wazo_dird/plugins/source_result.py
```diff
@@ -104,25 +110,24 @@ def __repr__(self) -> str:
...
     class SourceResult(_SourceResult):
         source = source_name
         backend = source_backend
         _unique_column = unique_column
         _format_columns = format_columns
-
-    SourceResult.is_deletable = is_deletable
-    SourceResult.is_personal = is_personal
+        is_deletable = _is_deletable
+        is_personal = _is_personal
```
fblackburn1, line 131: Why is the benefit to use intermediate variable `_is_delettable` instead of only use `is_deletable` (same as `source_name`)

integration_tests/suite/test_phonebook_http.py
```diff
@@ -27,10 +32,20 @@
     VALID_TOKEN_SUB_TENANT,
 )
 
+logger = logging.getLogger(__name__)
+
+
+class DirdClient(_DirdClient):
+    phonebook_source: Command
```
fblackburn1, line 39: same as above

fblackburn1:
Doc is missing
- openapi (api.yml)
- CHANGELOG.md

fblackburn1 (changes requested):
Doc is missing
- openapi (api.yml)
- CHANGELOG.md

fblackburn1 (approved):

PR #171, opened 2023-08-01
Wazo 3271 phonebook source

DrPyser:
https://wazo-dev.atlassian.net/browse/WAZO-3271

### DIFF
--- a/integration_tests/suite/test_phonebook_http.py
+++ b/integration_tests/suite/test_phonebook_http.py
@@ -23,7 +23,6 @@ from hamcrest import (
     instance_of,
     not_,
     only_contains,
-    starts_with,
 )
 from wazo_dird_client import Client as DirdClient
 from wazo_test_helpers.hamcrest.raises import raises
@@ -491,11 +490,9 @@ class TestGetContacts(BasePhonebookCRUDTestCase):
                 has_entries(
                     total=len(self.contacts),
                     filtered=len(self.contacts),
-                    items=all_of(
-                        contains_exactly(
-                            has_entries(firstname='Contact 0'),
-                            has_entries(firstname='Contact 1'),
-                        ),
+                    items=contains_exactly(
+                        has_entries(firstname='Contact 0'),
+                        has_entries(firstname='Contact 1'),
                     ),
                 ),
             )
@@ -511,6 +508,8 @@ class TestGetContacts(BasePhonebookCRUDTestCase):
             assert_that(
                 body,
                 has_entries(
+                    total=len(self.contacts),
+                    filtered=len(self.contacts),
                     items=contains_exactly(
                         has_entries(firstname='Contact 2'),
                         has_entries(firstname='Contact 3'),
@@ -540,12 +539,9 @@ class TestGetContacts(BasePhonebookCRUDTestCase):
                 has_entries(
                     total=len(self.contacts),
                     filtered=1,
-                    items=all_of(
-                        has_length(1),
-                        only_contains(
-                            has_entries(
-                                firstname=starts_with('Contact 4'),
-                            ),
+                    items=contains_exactly(
+                        has_entries(
+                            firstname='Contact 4',
                         ),
                     ),
                 ),
@@ -601,7 +597,6 @@ class TestPluginLookup(BasePhonebookCRUDTestCase):
         assert len(cls.contacts) == cls.contact_count
 
         source_body = {
-            # 'auth': {'host': 'auth', 'port': 9497, 'prefix': None, 'https': False},
             'first_matched_columns': ['number', 'firstname', 'lastname'],
             'format_columns': {
                 'phone': '{number}',
@@ -673,8 +668,7 @@ class TestPluginLookup(BasePhonebookCRUDTestCase):
             ),
         )
 
-        # bound response time to a linear function of the number of contacts
-        max_time = min(len(self.contacts) * 0.01, 5)
+        max_time = 2
         assert (
             timing['elapsed'] < max_time
         ), f'Lookup took too long {timing["elapsed"]} max {max_time}'
@@ -754,9 +748,7 @@ class TestGetContactsLoad(BasePhonebookCRUDTestCase):
             },
         ) as source:
             limit = 100
-            contact_ids: set[str] = set()
-            for i in range(5000 // limit):
-                logger.debug("request %d", i)
+            for i in range(self.num_contacts // limit):
                 with timed() as timing:
                     body = client.backends.list_contacts_from_source(
                         backend='phonebook',
@@ -782,12 +774,10 @@ class TestGetContactsLoad(BasePhonebookCRUDTestCase):
                         ),
                     ),
                 )
-                assert timing['elapsed'] < min(limit * 0.01, 5)
-                contact_ids.update(contact['id'] for contact in body['items'])
-
-            assert_that(
-                contact_ids, equal_to(set(contact['id'] for contact in self.contacts))
-            )
+                max_time = 2
+                assert (
+                    timing['elapsed'] < max_time
+                ), f'Get paginated took too long: {timing["elapsed"]} max {max_time}'
 
     def test_get_filtered(self):
         client = self.get_client(VALID_TOKEN_MAIN_TENANT)
@@ -823,16 +813,7 @@ class TestGetContactsLoad(BasePhonebookCRUDTestCase):
                     ),
                 ),
             )
-            assert timing['elapsed'] < min(self.num_contacts * 0.01, 5)
-            contact_ids = set(contact['id'] for contact in body['items'])
-
-            assert_that(
-                contact_ids,
-                equal_to(
-                    set(
-                        contact['id']
-                        for contact in self.contacts
-                        if 'Contact 4' in contact.get('firstname', '')
-                    )
-                ),
-            )
+            max_time = 2
+            assert (
+                timing['elapsed'] < max_time
+            ), f'Get filtered took too long: {timing["elapsed"]} max {max_time}'
--- a/integration_tests/suite/test_phonebook_http.py
+++ b/integration_tests/suite/test_phonebook_http.py
@@ -2,26 +2,36 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
-from contextlib import contextmanager
+from contextlib import ExitStack, contextmanager
+from typing import Iterator, TypedDict
 from unittest.mock import ANY
 from uuid import uuid4
 
 import requests
 from hamcrest import (
+    all_of,
     assert_that,
     calling,
     contains,
     contains_inanyorder,
     equal_to,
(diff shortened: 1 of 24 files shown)


DrPyser:
recheck

DrPyser:
recheck

wazo_dird/plugins/phonebook_backend/schemas.py
```diff
@@ -26,3 +32,19 @@ class ListSchema(_ListSchema):
...
 source_schema = SourceSchema()
 list_schema = ListSchema()
+
+
+class CountParams(TypedDict):
+    search: str | None
+
+
+class ContactListSchema(_ListSchema):
+    searchable_columns = ['uuid', 'firstname', 'lastname']
```
sduthil, line 42: Why is `uuid` searchable?

wazo_dird/plugins/phonebook_backend/schemas.py
```diff
@@ -26,3 +32,19 @@ class ListSchema(_ListSchema):
...
+    sort_columns = ['firstname', 'lastname']
```
sduthil, line 43: * `GET /phonebooks/uuid/contacts` allows `order` = `firstname` or `lastname` or `name`. Any idea why `name`? Can we homogenize the two API?

wazo_dird/plugins/phonebook_backend/schemas.py
```diff
@@ -26,3 +32,19 @@ class ListSchema(_ListSchema):
...
+    default_sort_column = 'lastname'
+
+    def load_count(self, args: dict, **kwargs) -> CountParams:
+        return cast(CountParams, projection(self.load(args, **kwargs), ['search']))
```
sduthil, line 47: IMHO this logic should stay contained in `http.py`. The reason is that if we change the counting logic, we should not have to change the schema code. That's more an opinion open to discussion though :)

If we keep this logic in here, we should rename it somewhat, because my first reaction was "what does count have to do with the schema???"

integration_tests/test-requirements-for-tox.txt
```diff
@@ -11,19 +13,19 @@ sh
 
 # for database tests
 psycopg2-binary
-sqlalchemy==1.2.18
-unidecode==1.0.23
-cheroot==6.5.4
+sqlalchemy
```
sduthil, line 16: I don't understand why the versions should be unfrozen. We should keep the code tested as close to production as possible, no?

wazo_dird/plugins/phonebook_backend/http.py
```diff
@@ -36,3 +50,79 @@ def get(self, source_uuid):
...
+        visible_tenants = get_tenant_uuids(recurse=False)
+        list_params = contact_list_schema.load(request.args)
+        count_params = contact_list_schema.load_count(request.args)
+        source_config = cast(
+            PhonebookSourceInfo,
+            self._source_service.get(self.BACKEND, str(source_uuid), visible_tenants),
+        )
+        logger.debug("Found source (uuid=%s): %s", source_uuid, str(source_config))
+
+        # sources prior to WAZO-3267 might be missing 'phonebook_uuid' and will crash here
```
sduthil, line 81: We should add a migration script to remove the invalid sources and remove this comment.

integration_tests/suite/test_phonebook_http.py
```diff
@@ -358,3 +391,304 @@ def test_get_multi_tenant(self, sub, main):
...
+        cls.contacts, errors = cls.contact_crud.create_many(
+            [MAIN_TENANT],
+            PhonebookKey(uuid=cls.phonebook['uuid']),
+            [
+                {
+                    'firstname': f'Contact {i}',
+                    'lastname': 'McContact',
+                    'number': str(1000000000 + i),
+                }
+                for i in range(5000)
```
sduthil, line 420: Open to discussion: having a lot of payload can make debugging the tests much harder. I would prefer separating logic tests (with smaller payloads) from load tests (with larger payloads) that only do timing assertions.

integration_tests/suite/test_phonebook_http.py
```diff
@@ -358,3 +391,304 @@ def test_get_multi_tenant(self, sub, main):
...
+                    total=len(self.contacts),
+                    filtered=len(self.contacts),
+                    items=all_of(
+                        has_length(len(self.contacts)),
+                        only_contains(
+                            has_entries(
+                                firstname=instance_of(str),
+                                lastname=instance_of(str),
+                                number=instance_of(str),
+                                id=instance_of(str),
```
sduthil, line 457: This does not assert that we have all contacts: the test would pass if the listing returned 5000 times the same contact.

integration_tests/suite/test_phonebook_http.py
```diff
@@ -358,3 +391,304 @@ def test_get_multi_tenant(self, sub, main):
...
+                        total=len(self.contacts),
+                        filtered=len(self.contacts),
+                        items=all_of(
+                            has_length(100),
+                            only_contains(
+                                has_entries(
+                                    firstname=instance_of(str),
+                                    lastname=instance_of(str),
+                                    number=instance_of(str),
+                                    id=instance_of(str),
```
sduthil, line 496: This test does not assert that we have the correct page: the test would pass if pages were ordered randomly

wazo_dird/database/queries/phonebook.py
```diff
@@ -129,19 +133,51 @@ def _new_search_filter(self, pattern, columns) -> bool | ColumnElement:
...
+        if search
+        else True
+    )
+    return search_filter
+
+
 class PhonebookContactCRUD(BaseDAO):
+    def __init__(self, Session: scoped_session):
+        super().__init__(Session)
+
```
sduthil, line 154: Is this `__init__` useful?

wazo_dird/database/queries/phonebook.py
```diff
@@ -303,35 +335,61 @@ def _get_contact(
...
         self,
-        s: Session,
-        query: ColumnElement,
+        s: BaseSession,
         phonebook_key: PhonebookKey,
-        search: str | None,
-    ):
+        search: str | None = None,
+    ) -> Query:
         phonebook_filter = contact_phonebook_key_to_predicate(phonebook_key)
```
sduthil, line 357: A bit out of scope, but I feel we have two words for the same thing: `filter` and `predicate` refer to the same notion, right? Can we use only one word?

wazo_dird/database/queries/phonebook.py
```diff
@@ -303,35 +335,61 @@ def _get_contact(
...
+    def _query_contact_fields(
+        self,
+        s: BaseSession,
+        phonebook_key: PhonebookKey,
+        search: str | None = None,
+    ) -> Query:
+        phonebook_filter = contact_phonebook_key_to_predicate(phonebook_key)
+        search_filter = contact_search_filter(
+            phonebook_key=phonebook_key, search=search
+        )
```
sduthil, line 374: Compared to the old version, this introduces an unnecessary subquery. Can we avoid that?

integration_tests/suite/test_phonebook_http.py
```diff
@@ -358,3 +391,304 @@ def test_get_multi_tenant(self, sub, main):
...
+                                '1000000005',
+                            ),
+                        )
+                    ),
+                ),
+            ),
+        )
+
+        # bound response time to a linear function of the number of contacts
+        assert_that(timing['elapsed'] < len(self.contacts) * 0.01)
```
sduthil, line 671: Please use `assert` for simple assertions, and add an assertion message.

Also, this places a threshold at 5000 * 0.01 = 50 seconds, which is unacceptable performance IMHO.

integration_tests/suite/test_phonebook_http.py
```diff
@@ -358,3 +391,304 @@ def test_get_multi_tenant(self, sub, main):
...
+                    set(
+                        contact['id']
+                        for contact in self.contacts
+                        if 'Contact 4' in contact.get('firstname', '')
+                    )
+                ),
+            )
+
+
+def timed(func, *args, **kwargs):
```
sduthil, line 556: Suggestion (did not try it), not really sure if I like "leaking" the context variable `timing` out of the block:
```
@contextmanager
def timed():
    result = {}
    start = time.time()
    try:
        yield result
    finally:
        end = time.time()
        result['elapsed'] = end - start

...
with timed() as timing:
    my_function('blah')
assert timing['elapsed'] < 1
```

wazo_dird/utils.py
```diff
@@ -0,0 +1,5 @@
+from typing import Any
+
+
+def projection(m: dict[str, Any], keys: list[str], default=None) -> dict[str, Any]:
```
sduthil, line 4: IMHO This needs some unit testing and/or some docstring

sduthil (changes requested):
Thanks for all this :)

Another problem I had when testing manually: I could not use the API with root token + recurse = True.

integration_tests/test-requirements-for-tox.txt
DrPyser: some cleanup to do to deduplicate

integration_tests/suite/test_phonebook_http.py
DrPyser: Add fix ceiling

wazo_dird/database/queries/phonebook.py
DrPyser: Actually, there's a logical difference, though I admit I'm not sure how the ORM is dealing with it. The subquery selects the contacts that apply based on their fields, which the main query pulls in whole. So a contact is represented by multiple rows in the main query, and we want to filter _per group_, not individual rows(we want to exclude the contacts that have no matching fields, not the fields that do not match). I'm not sure it's possible to do that without a subquery?

wazo_dird/plugins/phonebook_backend/http.py
DrPyser: Add ticket in comment : https://wazo-dev.atlassian.net/browse/WAZO-3357

wazo_dird/database/queries/phonebook.py
pc-m: I personally do not have a problem with subqueries if it makes the query easier to understand and modify. Unless there's a known problem I would not rework this query at the moment.

sduthil:
Thanks Pascal, most of the issues are fixed.
I still have a few changes needed:
* The `recurse` option on the new API is not working for me (with a root token, listing contacts returns 404)
* There is a missing PR in wazo-dird-client
* On a less blocking note, the integration tests could be simplified (assertions and client URLs)

sduthil:
After discussion, there is no PR missing in wazo-dird-client, there is already a generic function for that.

sduthil (approved):

sduthil (approved):

sduthil:
recheck

PR #174, opened 2023-08-10
Mypy linting

DrPyser:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - id: mypy
+        language_version: "3.10"
+        args: [--explicit-package-bases,--config-file,pyproject.toml]
+        additional_dependencies:
+          [
+            "types-requests",
+            "sqlalchemy-stubs",
+            "types-pyYAML",
+            "types-pkg-resources",
+          ]
```
sopelj, line 25:
```suggestion
        additional_dependencies:
          - "sqlalchemy-stubs"
          - "types-pkg-resources"
          - "types-pyyaml"
          - "types-requests"
          - "types-setuptools"
        exclude: "build|broken-plugins/setup.py"
```

pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+follow_imports = "skip"
+ignore_missing_imports = true
+
+plugins = [
+    "sqlmypy"
+]
+exclude = [
+    "build",
+    "integration_tests/docker/broken-plugins/setup.py"
+]
```
sopelj, line 37:
```suggestion
```

.pre-commit-config.yaml
sopelj: I think the reason `exclude` doesn't work in the pyproject.toml is that pre-commit is the one finding the files and passing them to mypy. If you exclude them via pre-commit it seems to work.

pyproject.toml
DrPyser: I'd want to keep the exclude here too, because pre-commit is not the only way to use mypy. Editor integrations and manually running mypy(e.g. for more hands-on debugging or playing around with mypy options without the pre-commit layer) should benefit from the config too.

.pre-commit-config.yaml
DrPyser: Oh right, make sense!

pyproject.toml
sopelj: I suppose... I just hate redundancy. I don't see why it would be necessary to run it another way. I agree it would be better if we could centralize it here, but it doesn't seem to work.

DrPyser: I don't love redundancy, but I prefer flexibility in this case.

PR #177, opened 2023-08-14
remove swagger version in phonebook API plugin

ja0x:
why: redefines the global version found in API plugin,and breaks our reference documentation


fblackburn1 (approved):

ja0x:
recheck

fblackburn1 (approved):

PR #181, opened 2023-09-14
[WAZO-3372] [WAZO-3385] phonebook sync

DrPyser:
depends-on: https://github.com/wazo-platform/wazo-dird-client/pull/24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 23.14
+* Field `phonebook_uuid` of phonebook sources is now validated to point to a valid phonebook,
+  and the phonebook source is now deleted along with the phonebook it points to(delete cascade)
+* database migrations now ensures existing phonebook sources refer to a valid phonebook using the `phonebook_uuid` configuration attribute. Existing sources using a `phonebook_id` attribute or those with a `name` matching that of an existing phonebook are updated with a `phonebook_uuid` attribute corresponding to that phonebook's `uuid`. Those phonebook sources that cannot be resolved to a valid phonebook are deleted.
+
 ## 23.12
 
 * Added new endpoint to phonebook backend API:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Changelog
 
 ## 23.14
-* Field `phonebook_uuid` of phonebook sources is now validated to point to a valid phonebook,
-  and the phonebook source is now deleted along with the phonebook it points to(delete cascade)
-* database migrations now ensures existing phonebook sources refer to a valid phonebook using the `phonebook_uuid` configuration attribute. Existing sources using a `phonebook_id` attribute or those with a `name` matching that of an existing phonebook are updated with a `phonebook_uuid` attribute corresponding to that phonebook's `uuid`. Those phonebook sources that cannot be resolved to a valid phonebook are deleted.
+* Field `phonebook_uuid` of phonebook sources is now validated to depend on a valid phonebook
+* Phonebook sources are now deleted along with the phonebook they depend upon
+* Existing phonebook sources that cannot be resolved to a valid phonebook are deleted.
 
 ## 23.12
 
--- a/alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
+++ b/alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
@@ -1,8 +1,7 @@
-"""remove-orphaned-phonebook-sources
+"""remove orphaned phonebook sources
 
 Revision ID: 3b2bfef244e7
 Revises: bb2cd24f0500
-
 """
 
 from alembic import op
@@ -11,7 +10,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '3b2bfef244e7'
-down_revision = 'bb2cd24f0500'
+down_revision = 'a3a7212e29b8'
 
 phonebook_table = sa.table(
     'dird_phonebook',
@@ -25,48 +24,19 @@ source_table = sa.table(
     sa.column('extra_fields'),
     sa.column('tenant_uuid'),
     sa.column('backend'),
+    sa.column('phonebook_uuid'),
 )
 
 
 def upgrade():
-    phonebooks = {}
-    sources = {}
-
-    query = sa.select(
-        [
-            phonebook_table.c.uuid,
-            phonebook_table.c.tenant_uuid,
-        ]
+    query = source_table.delete().where(
+        sa.and_(
+            source_table.c.backend == 'phonebook',
+            source_table.c.phonebook_uuid.is_(None),
+        )
     )
-    for row in op.get_bind().execute(query):
-        phonebooks[str(row.uuid)] = str(row.tenant_uuid)
-
-    query = sa.select(
-        [
-            source_table.c.uuid,
-            source_table.c.extra_fields,
-            source_table.c.tenant_uuid,
-        ]
-    ).where(source_table.c.backend == 'phonebook')
-    for row in op.get_bind().execute(query):
-        phonebook_uuid = str(row.extra_fields.get('phonebook_uuid'))
-        sources[row.uuid] = {
-            'tenant_uuid': str(row.tenant_uuid),
-            'phonebook_uuid': phonebook_uuid,
-        }
-
-    to_delete = set()
-    for source_uuid, source_details in sources.items():
-        phonebook_uuid = source_details['phonebook_uuid']
-        if not phonebook_uuid:
-            to_delete.add(source_uuid)
-        elif phonebook_uuid not in phonebooks:
-            to_delete.add(source_uuid)
-        elif source_details['tenant_uuid'] != phonebooks[phonebook_uuid]:
-            to_delete.add(source_uuid)
-
-    query = source_table.delete().where(source_table.c.uuid.in_(to_delete))
-    op.get_bind().execute(query)
+    result = op.get_bind().execute(query)
+    print("rows deleted:", result.rowcount)
 
 
 def downgrade():
--- a/alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
+++ b/alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
@@ -24,7 +24,7 @@ source_table = sa.table(
     sa.column('uuid'),
     sa.column('extra_fields'),
     sa.column('tenant_uuid'),
-    sa.column('backend')
+    sa.column('backend'),
 )
 
 
@@ -65,9 +65,7 @@ def upgrade():
         elif source_details['tenant_uuid'] != phonebooks[phonebook_uuid]:
             to_delete.add(source_uuid)
 
-    query = source_table.delete().where(
-        source_table.c.uuid.in_(to_delete)
-    )
+    query = source_table.delete().where(source_table.c.uuid.in_(to_delete))
     op.get_bind().execute(query)
 
 
--- a/alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
+++ b/alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
@@ -41,7 +41,6 @@ def upgrade():
     for row in op.get_bind().execute(query):
         phonebooks[str(row.uuid)] = str(row.tenant_uuid)
 
-
     query = sa.select(
         [
             source_table.c.uuid,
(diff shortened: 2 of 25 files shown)


pc-m (approved):

wazo_dird/plugins/phonebook_backend/plugin.py
```diff
@@ -49,18 +53,38 @@ def load(self, dependencies: PhonebookViewDependencies):  # type: ignore[overrid
             resource_class_args=args,
         )
 
-
-class _Config(TypedDict):
-    name: str  # phonebook source name
-    tenant_uuid: str
+        def on_phonebook_delete_cascade(phonebook_uuid: str):
```
pc-m, line 56: This should happen in a db migration script and then the code should not allow this situation to happen again

wazo_dird/plugins/phonebook_backend/http.py
```diff
@@ -56,15 +104,33 @@ def delete(self, source_uuid):
     def get(self, source_uuid):
         return super().get(source_uuid)
 
+    def _prepare_source_update(self, source_data: dict) -> PhonebookSourceInfo:
+        assert 'phonebook_uuid' in source_data
```
pc-m, line 108: the assertion should be removed to avoid having to handle a key error and an assertion error depending on the execution context

pc-m (changes requested):

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
```diff
@@ -0,0 +1,59 @@
...
+    query = sa.select(
+        [
+            source_table.c.uuid,
+            source_table.c.extra_fields,
+        ]
+    ).where(source_table.c.backend == 'phonebook')
+
+    phonebook_uuid_map = {}
+    for row in op.get_bind().execute(query):
+        phonebook_uuid_map[row.uuid] = row.extra_fields['phonebook_uuid']
```
pc-m, line 45: phonebook sources configured before 23.10 will still have a `phonebook_id` in extra_fields.
I would probably add another step at the beginning of this script or in a previous migration to convert all phonebook_id to phonebook_uuid

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
```diff
@@ -0,0 +1,59 @@
...
+
+    for source_uuid, phonebook_uuid in phonebook_uuid_map.items():
+        query = (
+            source_table.update()
+            .where(
+                source_table.c.uuid == source_uuid,
+            )
+            .values(phonebook_uuid=phonebook_uuid)
```
pc-m, line 53: The `phonebook_uuid` should also be removed from `extra_fields` for consistency

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+        op.get_bind().execute(query)
+
+
+def downgrade():
+    op.drop_column('dird_source', 'phonebook_uuid')
```
pc-m, line 59: The rollback should add the phonebook_uuid back into extra_fields

pc-m (dismissed):

wazo_dird/plugins/phonebook_backend/plugin.py
```diff
@@ -138,16 +136,14 @@ def _get_phonebook_key(
...
-        raise InvalidConfigError(
-            f'sources/{self._source_name}/phonebook_name',
-            f'unknown phonebook {phonebook_name}',
-        )
+        else:
+            phonebook_name = config['name']
+            phonebooks = self._crud.list([tenant_uuid], search=phonebook_name)
+            for phonebook in phonebooks:
+                if phonebook['name'] == phonebook_name:
+                    return database.PhonebookKey(uuid=phonebook['uuid'])
```
sduthil, line 144: Did the source name allow matching with the phonebook name? If so, can we make a migration script to migrate sources with `name` but without `phonebook_uuid` to link them the correct `phonebook_uuid`, instead of simply deleting them?

DrPyser: Yeah, this could be done. There's no guarantee that the source `name` matches the `phonebook.name`, but it seems to be a supported convention.

wazo_dird/database/queries/phonebook.py
```diff
@@ -40,7 +40,6 @@ class PhonebookDict(TypedDict):
 
 
 def phonebook_key_to_filter(phonebook_key: PhonebookKey) -> ColumnElement:
-    assert 'id' in phonebook_key or 'uuid' in phonebook_key
```
DrPyser, line 43: This was discussed with @fblackburn1 in previous PR review.

I like using assert to signal expectations about data that are not explicit(e.g. when type annotations cannot account for them) and that should hold for the code to make sense. It's more a comment to communicate expectations to other programmers. And type checkers can use assert statements for type narrowing(https://mypy.readthedocs.io/en/stable/type_narrowing.html).
These assertions are not used to catch expected error conditions, but to provide sanity checks that clarify understanding of the system.

 There are no donwsides I can see to having asserts in production code, as long as they do not replace what should be expected error conditions(specific exception that should be raised and potentially handled).
 
If the assert is triggered in production, it signals the code is broken and needs to be redesigned around that situation. If the assert would have been triggered but has been optimized out in production, the code will usually fail in another obvious way(such as here, when accessing the 'id' or 'uuid' key). The assert is a shortcut to make clear and explicit the expectations of the code and intent of the programmer. Short and explicit is better than long/complex and implicit.

 Removing assertions without replacing them with a proper alternative simply removes information.

wazo_dird/plugins/phonebook_backend/http.py
DrPyser: https://github.com/wazo-platform/wazo-dird/pull/181#discussion_r1340229950
https://mm.wazo.community/wazo/pl/83pgontwpfgqdg5wod17e4c1xo

DrPyser:
note: @fblackburn1 suggested further simplification if we are "hardlinking" source and phonebooks at the db levels. Working to directly integrate the `phonebook_name` and `phonebook_description` at the SourceCRUD DAO level.

wazo_dird/plugins/phonebook_backend/plugin.py
xxbillouxx: @DrPyser ?  :-)

wazo_dird/plugins/phonebook_backend/http.py
xxbillouxx: @DrPyser so is it resolved ?

wazo_dird/plugins/phonebook_backend/tests/test_dird_phonebook_backend.py
```diff
@@ -1,15 +1,20 @@
...
-from unittest.mock import Mock, patch
+from unittest.mock import MagicMock, Mock, patch
 from unittest.mock import sentinel as s
 
 from hamcrest import assert_that, equal_to
 
 from ..plugin import PhonebookPlugin, make_result_class
 
 
+def mock_dict(prototype):
```
xxbillouxx, line 14: is it used ?

wazo_dird/plugins/phonebook_backend/plugin.py
```diff
@@ -88,8 +88,6 @@ def load(self, dependencies: Dependencies):
         searched_columns = config.get('searched_columns')
         first_matched_columns = config.get('first_matched_columns')
 
-        self._crud = database.PhonebookCRUD(Session)
```
xxbillouxx, line 91: How `_crud` is initalized ?

wazo_dird/plugins/phonebook_backend/plugin.py
```diff
@@ -72,6 +71,7 @@ class PhonebookPlugin(BaseSourcePlugin):
     _source_name: str
     _search_engine: database.PhonebookContactSearchEngine
     _SourceResult: type[SourceResult]
+    _source_service: _SourceService
```
xxbillouxx, line 74: is it used ?

xxbillouxx (commented):
Very big PR :-)

wazo_dird/plugins/phonebook_backend/tests/test_dird_phonebook_backend.py
DrPyser: Does not seem to be, removing it.

wazo_dird/plugins/phonebook_backend/plugin.py
DrPyser: Good question, the line was removed by mistake as `_crud` is still used

DrPyser: Worryingly, that was not caught by tests as the tests use patch mocking on the `_crud` when the implementation would use it.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 23.14
+* Field `phonebook_uuid` of phonebook sources is now validated to point to a valid phonebook,
+  and the phonebook source is now deleted along with the phonebook it points to(delete cascade)
```
sduthil, line 5:
```suggestion
* Field `phonebook_uuid` of phonebook sources is now validated to depend on a valid phonebook
* Phonebook sources are now deleted along with the phonebook they depend upon
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+* database migrations now ensures existing phonebook sources refer to a valid phonebook using the `phonebook_uuid` configuration attribute. Existing sources using a `phonebook_id` attribute or those with a `name` matching that of an existing phonebook are updated with a `phonebook_uuid` attribute corresponding to that phonebook's `uuid`. Those phonebook sources that cannot be resolved to a valid phonebook are deleted.
```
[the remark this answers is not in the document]
sduthil: The changelog only needs to explain behavior changes from an API consumer point of view. Technical details are in the code / git history.

alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
```diff
@@ -0,0 +1,43 @@
...
+
+def upgrade():
+    query = source_table.delete().where(
+        sa.and_(
+            source_table.c.backend == 'phonebook',
+            source_table.c.phonebook_uuid.is_(None),
+        )
+    )
+    result = op.get_bind().execute(query)
+    print("rows deleted:", result.rowcount)
```
sduthil, line 39:
```suggestion
```

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
```diff
@@ -0,0 +1,139 @@
...
+            try:
+                phonebook = phonebook_by_name[row.name]
+                phonebook_uuid_map[row.uuid] = phonebook.uuid
+            except KeyError:
+                print(
+                    f'could not map source {row.uuid} to a phonebook. Future migration might delete it.'
+                )
+                continue
+
+    print('source to phonebook: ', phonebook_uuid_map)
```
sduthil, line 97:
```suggestion
```

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
```diff
@@ -0,0 +1,139 @@
...
+    if phonebook_uuid_map:
```
sduthil, line 98: This condition is not necessary, as the query will do nothing with an empty dict.

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
```diff
@@ -0,0 +1,139 @@
...
+                extra_fields=(
+                    sa.cast(source_table.c.extra_fields, postgresql.JSONB)
+                    - 'phonebook_uuid'
+                ),
+            )
+        )
+        op.get_bind().execute(
+            phonebook_sources_update,
+            [
+                {'_source_uuid': uuid, '_phonebook_uuid': phonebook_uuid}
```
sduthil, line 115: Out of curiosity, why the leading `_`?

sduthil (changes requested):

alembic/versions/3b2bfef244e7_remove_orphaned_phonebook_sources.py
DrPyser: Accumulate in which way? Because migrations can be ran and logged multiple times on a system?

alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
DrPyser: Honestly not sure, I think I feared namespace clashes after seeing some logs.

sduthil (approved):

sduthil (approved):

PR #186, opened 2023-11-15
Fixed phonebook source migration

DrPyser:
when using the source name, was not limiting match to the same tenant scope, leading to sources matched to phonebooks not respecting the foreign key constraint and breaking the migration.

### DIFF
--- a/alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
+++ b/alembic/versions/a3a7212e29b8_source_add_phonebook_uuid.py
@@ -72,7 +72,7 @@ def upgrade():
         ]
     )
     phonebooks = list(op.get_bind().execute(phonebooks_query))
-    phonebook_by_name = {(p.name, p.tenant_uuid): p for p in phonebooks}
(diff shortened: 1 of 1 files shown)


pc-m:
@DrPyser this migration script was introduced in 23.14 and as already been executed.
Is it possible to have a new migration script that would repair migrations that have already been executed and leave the current version of the script as is?

DrPyser:
> @DrPyser this migration script was introduced in 23.14 and as already been executed. Is it possible to have a new migration script that would repair migrations that have already been executed and leave the current version of the script as is?

Well if any migration was on the failure path(phonebook matching the source name but in another tenant), the migration failed and did not go through until the administrator manually fixed the situation(e.g. by deleting or renaming the source, by deleting or renaming the phonebook).
If the migration did not encounter the failure path, all should be good: the phonebook sources will have been associated to a phonebook in the same tenant, or no matching phonebook would be found and the source would be deleted by the next migration.

So I don't think a new migration could fix anything that wasn't already necessary to do to fix this migration.

By fixing this migration in place though(backporting the fix if deemed valuable) any new attempts to migrate would avoid the failure path without requiring manual actions.

pc-m (approved):

PR #192, opened 2024-01-31
plugins.phonebook_backend: fixed PhonebookSource definition

DrPyser:
phonebook openapi schema PhonebookSource definition has response-only attributes that must be set to `readOnly`, in accordance with actual marshmallow schemas.


fblackburn1 (approved):

fblackburn1:
(force push to stop zuul gate - will be merged after freeze)

pc-m:
This change is already in master

PR #199, opened 2024-06-14
Adding live debug logs

AkshitaPatel:
Depends-On: https://github.com/wazo-platform/wazo-dird-client/pull/29
Creating a patch API to allow the change of debug setting without restarting


jfbourque:
recheck

AkshitaPatel:
recheck

AkshitaPatel:
recheck

requirements.txt
```diff
@@ -12,6 +12,7 @@ flask==1.1.2
 graphene==3.2.1
 itsdangerous==1.1.0  # from flask
 jinja2==2.11.3  # from flask
+jsonpatch==1.21
```
pc-m, line 15: When adding a dependency to the `requirements.txt` file you also need to add it to the `debian/control` file which is used by the debian packaging.
You can look at wazo-auth `debian/control` as an example

integration_tests/suite/test_config.py
```diff
@@ -66,3 +71,32 @@ def _not_return_503():
...
+                'op': 'replace',
+                'path': '/debug',
+                'value': False,
+            }
+        ]
+
+        dird = self.make_dird(VALID_TOKEN_MAIN_TENANT)
+
+        debug_true_patched_config = dird.config.patch(debug_true_config)
+        debug_true_config = dird.config.get()
```
pc-m, line 95: You should not reuse the variable name `debug_true_config` here. It make the test confusing.
At the begining of the test the value is
```python
[
            {
                'op': 'replace',
                'path': '/debug',
                'value': True,
            }
        ]
```
and at the end it's
```python
{
    ...
     "debug": True,
     ...
 }
 ```
 You could name it `debug_true_updated_config`

Comment (pc-m) [reply integration_tests/suite/test_config.py:95]: Same comment for `debug_false_config`

File: wazo_dird/plugins/config_service/plugin.py [hunk 3]
```diff
@@ -140,16 +141,39 @@ def load(self, dependencies):
...
         self._config = config
         self._controller = controller
 
         self._bus.subscribe(TenantCreatedEvent.name, self._on_new_tenant_event)
 
-    def get_current_config(self):
+    def get_config(self):
         return self._config
 
+    def update_config(self, config: dict) -> None:
```
pc-m, line 158: Since you're introducing typing in wazo-dird you should add a check in the `.pre-commit-config.yaml` file to be sure your typing is correct and that it stays that way.

pc-m (changes requested):
Works great, I've added some inline comments to address but nothing major. Good Job :+1:

requirements.txt
AkshitaPatel: Thank you, It is done.

integration_tests/suite/test_config.py
AkshitaPatel: Thank you, It is done.

wazo_dird/plugins/config_service/plugin.py
AkshitaPatel: Thank you, It is done.

pc-m (approved):

PR #211, opened 2024-12-02
WAZO 3406 fix phonebook contact import

DrPyser:


wazo_dird/plugins/phonebook_service/plugin.py
```diff
@@ -161,19 +165,33 @@ def import_contacts(
...
+            return [], errors
 
         created, failed = self._contact_crud.create_many(
-            visible_tenants, phonebook_key, to_add
+            visible_tenants, phonebook_key, [contact for _, contact in to_add]
         )
 
-        return created, failed + errors
+        errors += failed
+        return created, errors
```
rbienvenu0, line 194: At this point, `errors` has to be empty (it would have been returned l.187 otherwise)
```suggestion
        return created, failed
```

wazo_dird/plugins/phonebook/http.py
```diff
@@ -148,24 +149,59 @@ def post(self, phonebook_uuid: UUID):
...
         )
+        if not created:
+            raise PhonebookContactImportAPIError(
+                message='failed to create contacts',
+                error_id='phonebook-contact-import-bad-contacts',
+                status_code=400,
+                details={'errors': failed},
+            )
 
         return {'created': created, 'failed': failed}
```
rbienvenu0, line 206: Since errors are raised during evalution/addition of the contacts, the `failed` field should always be empty now, no?
(and therefore could be removed)

rbienvenu0 (changes requested):

wazo_dird/plugins/phonebook_service/plugin.py
DrPyser: Yes. Since we're not changing the API interface yet(which still returns `{"created": ..., "failed": ...}`), we're keeping the same interface at this level.

wazo_dird/plugins/phonebook/http.py
rbienvenu0: Fair enough. Is there already a Jira ticket to capture the change? Should we create one?

wazo_dird/plugins/phonebook_service/plugin.py
rbienvenu0: This comment is not about changing the API interface, just skipping the line `errors += failed` and returning `failed` directly instead, because at that point `errors` should always be empty.

wazo_dird/plugins/phonebook/http.py
DrPyser: Yeah, I'll create a ticket

DrPyser: https://wazo-dev.atlassian.net/browse/WAZO-3977

wazo_dird/plugins/phonebook/http.py
```diff
@@ -148,24 +149,59 @@ def post(self, phonebook_uuid: UUID):
 
         charset = request.mimetype_params.get('charset', 'utf-8')
         raw_data = cast(bytes, request.data)
+        logger.debug('len(raw_data)=%d', len(raw_data))
```
rbienvenu0, line 152: Does keeping this debug line have value? Or was it to help during development?

rbienvenu0 (approved):

PR #218, opened 2025-01-10
remove no-op transaction with psycopg2

fblackburn1:
- psycopg <2.9 (bullseye): when autocommit=True, the context manager is a noop
- psycopg >= 2.9 (bookworm): context manager always open a transaction
  and it's imcompatible with CREATE DATABASE


rbienvenu0 (approved):

fblackburn1:
recheck (issue with zuul DNS)

PR #221, opened 2025-03-21
[WAZO-4094] Link users to tenant

bloom1:


alembic/versions/2adc8aff56ea_add_tenant_uuid_to_dird_user.py
```diff
@@ -0,0 +1,40 @@
...
+    op.create_index(
+        f'{user_table_name}__idx__tenant_uuid',
+        user_table_name,
+        ['tenant_uuid'],
+    )
+
+
+def downgrade():
+    op.drop_column(user_table_name, 'tenant_uuid')
+    op.drop_index(f'{user_table_name}__idx__tenant_uuid')
```
rbienvenu0, line 40: To keep it explicit, dropping the foreign key should be done before dropping the column.
`op.drop_column()` already drops the foreign key associated with the column if there's one. leading to `op.drop_index()` trying to get rid of a non-existing resource, and raising an error.

pc-m (dismissed):

pc-m (approved):

alembic/versions/2adc8aff56ea_add_tenant_uuid_to_dird_user.py
bloom1: Fixed, thank you

PR #224, opened 2025-04-02
WAZO-4100 migrate existing numbers to e164

rbienvenu0:
WAZO-4100


alembic/versions/9569b2f8e588_add_e164_formatted_number_for_contacts.py
```diff
@@ -0,0 +1,135 @@
...
+        contact_field_table.join(
+            contact_table,
+            contact_field_table.c.contact_uuid == contact_table.c.uuid,
+        )
+        .join(
+            phonebook_table,
+            contact_table.c.phonebook_uuid == phonebook_table.c.uuid,
+        )
+        .select()
+        .where(contact_field_table.c.name == 'number')
```
pc-m, line 70: What about columns name "fax", "mobile", "numéro", "num", "exten", etc
We are missing a concept on phonebooks and personal contacts that would allow dird to know which column is a number and which column is not.

contribs/docker/wazo-dird-db.Dockerfile
```diff
@@ -7,6 +7,7 @@ ENV ALEMBIC_DB_URI=postgresql://wazo-dird:Secr7t@localhost/wazo-dird
 
 RUN true \
     && python3 setup.py install \
+    && pip install phonenumbers==8.12.1\
```
pc-m, line 10: The `debian/control` also needs this dependency

pc-m (changes requested):
This PR should not be merged if the following one that does the formatting is done at the same time.

As explained in the inline comment we need a way to know which column contains a number. A user might use the column `number` and `mobile` in its application, that allows the developer of that application to do something specific about those columns in the context of that application but the dird service itself cannot make that kind of guess.

contribs/docker/wazo-dird-db.Dockerfile
rbienvenu0: Updated

PR #226, opened 2025-04-04
WAZO-4102 multiple reverse

rbienvenu0:


wazo_dird/plugins/graphql/plugin.py
```diff
@@ -36,7 +39,23 @@ def _is_authorized(self, info: ResolveInfo, token_id: str) -> bool:
...
         required_acl = f'dird.graphql.{root_field}'
         try:
-            token_is_valid = self._auth_client.token.is_valid(token_id, required_acl)
+            tenant = Tenant.autodetect(self._auth_client)
+        except AuthServerUnreachable as e:
+            host = self._auth_config['host']
+            port = self._auth_config['port']
+            raise graphql_error_from_api_exception(AuthServerUnreachable(host, port, e))
+        except APIException:
+            return False
```
DrPyser, line 48: Would this be an error condition, or a normal condition?
If error, would be good to log something know what the error is about.
If normal, can we expect a specific exception instead of a generic `APIException`?

DrPyser (approved):
All good, good job!

PR #227, opened 2025-04-18
WAZO-3954 replace slashes in tenants' name

rbienvenu0:
Why: some APIs rely on the tenant's name and having a slash in it breaks the URL

### DIFF
--- a/integration_tests/suite/test_displays.py
+++ b/integration_tests/suite/test_displays.py
@@ -1,4 +1,4 @@
-# Copyright 2019-2023 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2019-2025 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from contextlib import contextmanager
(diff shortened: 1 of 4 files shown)


integration_tests/suite/test_displays.py
```diff
@@ -252,6 +252,24 @@ def test_valid_post(self):
...
 
+    def test_slash_in_name(self):
+        body = {
+            'name': 'display/test',
+            'columns': [
+                {'field': 'fn', 'title': 'Firstname', 'default': ''},
+            ],
+        }
+        with self.create(self.client, body) as display:
+            print(display)
```
bloom1, line 263: You forgot a print in tests :smile:

wazo_dird/database/models.py
```diff
@@ -307,7 +315,7 @@ def name(self, value):
                 self.uuid,
             )
         else:
-            self._name = value
+            self._name = value.replace('/', '_')
```
bloom1, line 318: You're missing tests for `Source` ­— you added the tests for the displays, but sources should also be tested. There is code testing the source creation in `integration_tests/suite/test_auto_create.py`

bloom1 (changes requested):
Do you think we should also migrate the source and display names?

integration_tests/suite/test_displays.py
rbienvenu0: :sweat_smile:

wazo_dird/database/models.py
rbienvenu0: Updated

integration_tests/suite/test_displays.py
rbienvenu0: Updated

bloom1:
recheck

bloom1:
recheck

rbienvenu0:
recheck

ahogui12:
recheck

bloom1:
recheck

bloom1:
recheck

bloom1:
Superseded by https://github.com/wazo-platform/wazo-dird/pull/229

PR #253, opened 2026-06-16
ITEM-390 WAZO 2117 configured reverse timeout not effective

DrPyser:
## suggested test protocol

- [x] prior: verify that changing timeout option in profile through API is not effective (e.g. set to 0.01 to ensure reverse lookup should timeout systematically)
- [x] prior: verify that database has top-level `timeout` keys in default profile with `select name, config from dird_profile_service join dird_profile on dird_profile.uuid = dird_profile_service.profile_uuid ;`
- [x] mount PR 
  - [x] run database migrations (`xivo-upgrade-db`); verify that database now has config updated  with `timeout` in `options` sub-object
  - [x] change default profile timeout through API and verify that new value is effective
  - [ ] setting timeout to 0 or `null` yields error
  - [x] same works for lookup

---

> [!NOTE]
> **Medium Risk**
> Changes directory lookup/reverse timing and rejects previously accepted profile JSON shapes; deploy requires DB migration for existing timeout values.
> 
> **Overview**
> Fixes profile **reverse** (and **lookup**) timeouts so `options.timeout` is actually applied instead of a ignored top-level `timeout` field.
> 
> **Runtime:** reverse and lookup read timeout from `services.<name>.options.timeout`. Reverse always passes a timeout to `as_completed` (default **1s** when unset), cancels pending futures on timeout or early success, and logs timeouts at warning level. Auto-created tenant profiles now set `options: { timeout: 0.5 }` for reverse.
> 
> **API & data:** `POST`/`PUT` profile payloads strictly validate each service block and `options` (unknown keys → **400**; `timeout` must be **> 0** when set). An Alembic migration moves existing `dird_profile_service.config.timeout` into `config.options.timeout`.
> 
> **Tests:** unit tests for schemas and timeout behaviour, plus integration coverage with a slow CSV-WS source.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45da2b093dd25e457a4b5827d7dfe61ea6ee6d46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wazo_dird/plugins/lookup_service/tests/test_lookup.py
+++ b/wazo_dird/plugins/lookup_service/tests/test_lookup.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2023 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2014-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import unittest
@@ -6,7 +6,7 @@ from unittest.mock import Mock, patch, sentinel
 
 from hamcrest import assert_that, equal_to, none, not_
 
-from ..plugin import LookupServicePlugin
+from ..plugin import LookupServicePlugin, _LookupService
 
 
 class TestLookupServicePlugin(unittest.TestCase):
@@ -75,3 +75,24 @@ class TestLookupServicePlugin(unittest.TestCase):
         plugin.unload()
 
         MockedLookupService.return_value.stop.assert_called_once_with()
+
+
+def _make_lookup_service() -> _LookupService:
+    source_manager = Mock()
+    source_manager.get.return_value = None
+    return _LookupService(config={}, source_manager=source_manager, controller=Mock())
+
+
+class TestLookupNullOptions(unittest.TestCase):
+    @patch('wazo_dird.plugins.lookup_service.plugin.wait')
+    def test_null_options_does_not_crash(self, mock_wait):
+        mock_wait.return_value = ([], [])
+        service = _make_lookup_service()
+        profile = {
+            'name': 'test',
+            'services': {'lookup': {'sources': [], 'options': None}},
+        }
+
+        service.lookup(profile, 'tenant', 'alice', 'user-uuid')
+
+        mock_wait.assert_called_once()
--- a/wazo_dird/plugins/lookup_service/tests/test_lookup.py
+++ b/wazo_dird/plugins/lookup_service/tests/test_lookup.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import unittest
+from concurrent.futures import ALL_COMPLETED
 from unittest.mock import Mock, patch, sentinel
 
 from hamcrest import assert_that, equal_to, none, not_
@@ -95,4 +96,4 @@ class TestLookupNullOptions(unittest.TestCase):
 
         service.lookup(profile, 'tenant', 'alice', 'user-uuid')
 
-        mock_wait.assert_called_once()
+        mock_wait.assert_called_once_with([], return_when=ALL_COMPLETED)
--- a/wazo_dird/plugins/reverse_service/plugin.py
+++ b/wazo_dird/plugins/reverse_service/plugin.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2023 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2015-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
@@ -64,7 +64,7 @@ class _ReverseService(helpers.BaseService):
 
         params = {}
         service_config = self.get_service_config(profile_config)
-        timeout = service_config.get('timeout')
+        timeout = service_config.get('options', {}).get('timeout', 1)
         if timeout:
             params['timeout'] = timeout
 
@@ -109,7 +109,7 @@ class _ReverseService(helpers.BaseService):
 
         params = {}
         service_config = self.get_service_config(profile_config)
-        timeout = service_config.get('timeout')
+        timeout = service_config.get('options', {}).get('timeout', 1)
         if timeout:
             params['timeout'] = timeout
 
--- a/wazo_dird/plugins/reverse_service/plugin.py
+++ b/wazo_dird/plugins/reverse_service/plugin.py
@@ -62,15 +62,12 @@ class _ReverseService(helpers.BaseService):
             args['xivo_user_uuid'] = user_uuid
             futures.append(self._async_reverse_many(source, extens, args))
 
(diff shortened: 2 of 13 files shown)


wazo_dird/plugins/reverse_service/plugin.py
```diff
@@ -109,18 +130,27 @@ def reverse(
 
         params = {}
         service_config = self.get_service_config(profile_config)
-        timeout = service_config.get('timeout')
+        timeout = service_config.get('options', {}).get('timeout', 1)
         if timeout:
```
ja0x:
```suggestion
        if timeout is not None:
```
Is a value of 0 (zero) allowed to disable timeout? If not, only not having the option disables it?

wazo_dird/plugins/reverse_service/plugin.py
```diff
@@ -64,21 +64,42 @@ def reverse_many(
...
```
ja0x:
```suggestion
        if timeout is not None:
```
Same as next other timeout comment

wazo_dird/plugins/reverse_service/plugin.py
```diff
@@ -109,18 +130,27 @@ def reverse(
 
         params = {}
         service_config = self.get_service_config(profile_config)
-        timeout = service_config.get('timeout')
+        timeout = service_config.get('options', {}).get('timeout', 1)
```
ja0x: This also breaks the default top-level timeout value. In the `_auto_create_profile` we set the timeout value outside the `options` key, which is now dead code
```
'services': {
                'lookup': {'sources': sources},
                'favorites': {'sources': sources},
                'reverse': {'sources': sources, 'timeout': 0.5},
            },
```
Either remove the default here and let the `reverse_service` default explicitely take predence or move it under `{..., 'options': {'timeout': 0.5}}`

ja0x:
```suggestion
        timeout = (service_config.get('options') or {}).get('timeout', 1)
```
This forces the default if `options` key is intentionally/unintentionally nulled

wazo_dird/plugins/reverse_service/plugin.py
```diff
@@ -64,21 +64,42 @@ def reverse_many(
...
```
ja0x:
```suggestion
        timeout = (service_config.get('options') or {}).get('timeout', 1)
```

wazo_dird/plugins/reverse_service/plugin.py
```diff
@@ -109,18 +130,27 @@ def reverse(
...
             for future in as_completed(futures, **params):
-                if future.result() is not None:
-                    for other_future in futures:
-                        other_future.cancel()
-                    return future.result()
+                if result := future.result():
+                    pending_futures = [f for f in futures if not f.done()]
+                    cancelled_futures = sum(1 for f in pending_futures if f.cancel())
+                    logger.debug(
+                        'Cancelled %d/%d pending reverse lookup tasks',
```
ja0x, line 137: Watch out about the logging line, the futures are indeed being cancelled, but already scheduled (on workers) task will still run to completion. The comment could be slightly misleading

(Applies to all "cancelled X/X" logs)

ja0x (changes requested):
The fix works locally, nice work. A few points on default values + task cancellation. 

Subsequent to this PR:
Reading on `threadpools`, once running, we cannot cancel a task until completion easily/cleanly without hacking internals. Right now we make sure the reverse lookup returns fast in case of timeout, but the resources are still consumed. A later fix for this would be to bound the request's timeout directly. According to Claude, the CSV backend already implements it (requests.get(..., timeout=...). External third party providers (office, google) do not yet implement a timeout, something we could add to also release resources on a lookup timeout

wazo_dird/plugins/reverse_service/plugin.py
DrPyser: well a timeout of 0 is useless (as_completed would timeout directly).

Right now the timeout is optional but with a default value. 

I'm not sure it's worth trying to avoid all possible corner cases given the existing code is completely broken on that regard. The default timeout is unlikely to ever be changed, and if it is I assume it will be by our recommendation.

wazo_dird/plugins/reverse_service/plugin.py
DrPyser: hmm yes that was the whole point. Probably missed in cherry-picking.

wazo_dird/plugins/reverse_service/plugin.py
DrPyser: If you look at the logic, we actually count which calls to Future.cancel returned true, which is what indicates which future got actually cancelled. https://docs.python.org/3.11/library/concurrent.futures.html#concurrent.futures.Future.cancel

In tests (load tests), we can see that most futures are not actually cancelled (`0/0` or `0/[1-9]`) unless the timeout is really small or the executor pool is forced into a bottleneck w.r.t http worker threads.

wazo_dird/plugins/reverse_service/plugin.py
ja0x: Argh, good catch. 0 is immediate expiry, thought it was same as none

wazo_dird/plugins/reverse_service/plugin.py
ja0x: Right, futures being currently processed will noop on cancel; we just want to know actually cancelled futures, got it

wazo_dird/plugins/lookup_service/tests/test_lookup.py
```diff
@@ -75,3 +75,24 @@ def test_that_unload_stops_the_services(self, MockedLookupService):
...
+        mock_wait.return_value = ([], [])
+        service = _make_lookup_service()
+        profile = {
+            'name': 'test',
+            'services': {'lookup': {'sources': [], 'options': None}},
+        }
+
+        service.lookup(profile, 'tenant', 'alice', 'user-uuid')
+
+        mock_wait.assert_called_once()
```
ja0x: To strenghten the test, should the call args also be tested for the default value?

wazo_dird/plugins/reverse_service/tests/test_reverse.py
```diff
@@ -0,0 +1,174 @@
...
+    'name': 'test',
+    'services': {'reverse': {'sources': [{'uuid': _SOURCE_UUID}]}},
+}
+
+_PROFILE_WITH_TIMEOUT = {
+    'name': 'test',
+    'services': {'reverse': {'sources': [], 'options': {'timeout': 0.5}}},
+}
+
+# timeout outside options is stripped by the profile schema (EXCLUDE unknown fields)
```
ja0x:
```suggestion
```
This comment is an artifact from previous version; schema is now RAISE instead of EXCLUDE, should probably be removed

ja0x (dismissed):

PR #266, opened 2026-07-30
Use nginx for internal wazo-auth requests

fblackburn1:
Points requests to wazo-auth at the nginx internal entry point
(`localhost:80`) instead of the wazo-auth process port, in both the Python
defaults and the shipped `config.yml` (the YAML is loaded on top of the Python
defaults, so both must change).

The `prefix` override is dropped rather than set: `wazo-auth-client` already
defaults to `/api/auth`.

Depends on:

* wazo-platform/wazo-nginx#23 — the internal entry point itself
* wazo-platform/wazo-auth#449 — the wazo-auth location served on it

Both must be deployed before this merges, otherwise internal auth calls 404.
Ticket: TASK-504

Notes:

* Integration test assets now pin `port: 9497` and `prefix: null` explicitly:
  their stacks have no nginx, and those values used to come from the defaults
  changed here.
* Directory sources store their own auth config: the templates now point at
  nginx and alembic revision `4f2c91ab07de` migrates the existing ones, leaving
  sources that point elsewhere untouched. Tested against a scratch postgres.
* The `prefix` is set explicitly to `/api/auth` there, since it is stored data.

---

> [!NOTE]
> **Medium Risk**
> Changes how all internal auth traffic is routed and migrates stored source configs; requires nginx/wazo-auth changes to be deployed first or auth calls will 404.
> 
> **Overview**
> Routes internal **wazo-auth** traffic through nginx (`localhost:80` + `/api/auth`) instead of the direct process port `9497`.
> 
> Defaults in `config.py` and `config.yml` now use port `80` and drop the explicit `prefix: null` (the client already defaults to `/api/auth`). Auto-created source bodies and API docs are updated to match. An alembic migration rewrites existing `dird_source` auth configs that pointed at `localhost:9497`, leaving custom hosts/ports untouched.
> 
> The remote `wazo_users` template now targets port `443` with HTTPS. Integration test assets pin the old direct-auth settings because those stacks have no nginx.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 913cabdeafb55e3f9cdafdacfa9c4ba809123457. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/alembic/versions/4f2c91ab07de_migrate_source_auth_to_nginx.py
+++ b/alembic/versions/4f2c91ab07de_migrate_source_auth_to_nginx.py
@@ -1,0 +1,47 @@
+"""migrate_source_auth_to_nginx
+
+Revision ID: 4f2c91ab07de
+Revises: 5a67556fbbf1
+
+"""
+
+import sqlalchemy as sa
+
+# alembic exposes op as a runtime proxy that mypy cannot see statically
+from alembic import op  # type: ignore[attr-defined]
+
+# revision identifiers, used by Alembic.
+revision = '4f2c91ab07de'
+down_revision = '5a67556fbbf1'
+
+_table = sa.table(
+    'dird_source',
+    sa.column('uuid', sa.String),
+    sa.column('extra_fields', sa.JSON),
+)
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    rows = conn.execute(sa.select(_table)).fetchall()
+    for row in rows:
+        extra_fields = row.extra_fields
+        if not isinstance(extra_fields, dict):
+            continue
+        auth = extra_fields.get('auth')
+        if not isinstance(auth, dict):
+            continue
+        # sources pointing elsewhere are configured by the administrator
+        if auth.get('host') != 'localhost' or auth.get('port') != 9497:
+            continue
+        auth['port'] = 80
+        auth['prefix'] = '/api/auth'
+        conn.execute(
+            sa.update(_table)
+            .where(_table.c.uuid == row.uuid)
+            .values(extra_fields=extra_fields)
+        )
+
+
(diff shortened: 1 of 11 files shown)


alembic/versions/4f2c91ab07de_migrate_source_auth_to_nginx.py
```diff
@@ -0,0 +1,47 @@
...
+        if not isinstance(extra_fields, dict):
+            continue
+        auth = extra_fields.get('auth')
+        if not isinstance(auth, dict):
+            continue
+        # sources pointing elsewhere are configured by the administrator
+        if auth.get('host') != 'localhost' or auth.get('port') != 9497:
+            continue
+        auth['port'] = 80
+        auth['prefix'] = '/api/auth'
```
ja0x, line 38:
```suggestion
        auth['port'] = 80
        auth['prefix'] = '/api/auth'
        auth['https'] = False
        auth['verify_certificates'] = False
```
Should the migration script also set `https` / `verify_certificates` to the default value now too? Although, I'm pretty sure previous values need `https` at `False` to work

ja0x (changes requested):
`/etc/wazo-dird/templates.d/wazo_users.yml.jinja2` still used the old localhost:9497, it should probably also be migrated to localhost:80?

alembic/versions/4f2c91ab07de_migrate_source_auth_to_nginx.py
fblackburn1: exactly, `https` and `verify_certificate` was already at the right value to work

ja0x (approved):

PR #269, opened 2026-08-10
TASK-511-reuse-request-token

fblackburn1:
- **Reuse the request-scoped token instead of refetching it**
- **Add a wazo-auth request budget test**

---

> [!NOTE]
> **Medium Risk**
> Touches authentication-related request handling across many API surfaces; behavior should be equivalent but depends on request-scoped user/tenant population matching the old token metadata reads.
> 
> **Overview**
> **Stops redundant wazo-auth token fetches** by introducing shared `get_user_uuid()` (from `xivo.tenant_flask_helpers.user`) and using it across directory JSON routes (lookup, favorites, personal), personal CRUD/import, and GraphQL `me` resolution instead of `auth.client().token.get(...)`. The local `_get_calling_user_uuid()` helper in personal HTTP is removed in favor of the shared helper; GraphQL auth middleware now calls `Tenant.autodetect()` without passing the auth client.
> 
> **Adds integration coverage** (`TestTokenRequestCount`) that asserts how many `/0.1/token` GET/HEAD calls occur per request for lookup, reverse, favorites, personal, and GraphQL—encoding the rule that verify still hits auth once and a full token fetch happens at most once when tenant or user UUID is not already available from headers/path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d45ef39c4d990e3e91ff1dd4a01e118d9c658f16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/integration_tests/suite/test_auth_requests.py
+++ b/integration_tests/suite/test_auth_requests.py
@@ -1,0 +1,136 @@
+# Copyright 2026 The Wazo Authors  (see the AUTHORS file)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from hamcrest import assert_that, equal_to, has_key, has_length, not_
+from wazo_test_helpers.auth import AuthClient as MockAuthClient
+from wazo_test_helpers.auth import MockUserToken
+
+from .helpers.base import BaseDirdIntegrationTest
+from .helpers.config import new_csv_with_multiple_displays_config
+from .helpers.constants import MAIN_TENANT
+
+USER_UUID = 'my-user-uuid'
+
+
+class TestTokenRequestCount(BaseDirdIntegrationTest):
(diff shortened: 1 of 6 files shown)


DrPyser (commented):
Code-wise LGTM

DrPyser (commented):
lookup and favorites APIs see a visible behavior change, where tokens lacking user identity previous would result in 500, now in 401s.
Is there a concrete use case / normal scenario where this can occur? Or just invalid tokens that shouldn't exist?
If so, might be worth documenting in changelog?

DrPyser (commented):
Looks alright functionally, behave as documented!

DrPyser (approved):
Good for me, thanks!

PR #271, opened 2026-08-12
ITEM-594 PRODUCT-332: reuse the docker asset across integration test classes

DrPyser:
Moves the suite to `wazo_test_helpers.pytest_asset`, the shared plugin that
wazo-auth, wazo-chatd and others already use.
(wazo-agid and some others use the older `make_asset_fixture` with tests written for
pytest, so they need no marker and no asset class.) 

From the red bin [RB-51](https://app.notion.com/p/37baeaae6d2a810ca30ce9409779d6a0),
tracked as scaling card
[PRODUCT-332](https://app.notion.com/p/3baaeaae6d2a8109a0a9f30c0ad512af).

## The problem

`AssetLaunchingTestCase` starts the stack in `setUpClass` and stops it in
`tearDownClass`, so it destroyed and recreated the same asset back to back when
consecutive classes needed it: **153 starts for 24 assets**.

dird was the only project that still did this, because its test classes *were*
the asset classes. The plugin needs the two to be separate.

## What changes

- `DirdAssetLaunchingTestCase` and one subclass for each of the 24 assets own
  the lifecycle. `conftest.py` builds a session fixture for each of them and
  registers the plugin.
- `DirdAssetRunningTestCase` becomes a plain `unittest.TestCase`. A test class
  still names its stack with `asset = '<name>'`, and `__init_subclass__` gives
  it the matching `asset_cls` and the `usefixtures` marker.
- Seven small classmethods on the test base send `service_port`, `docker_exec`,
  `restart_service`, `stop_service`, `start_service`, `capture_logs` and
  `_run_cmd` to `asset_cls`, and `assets_root` is declared beside them. Those are
  every member of the asset class that the tests use, so the 69 calls that they
  inherited keep working and no test file changes.
- `test_database.py` and `test_dird_phonebook_backend.py` launched their stack
  from `setup_module`, which runs before the fixtures. They now use a
  module-scoped fixture that asks for the `database` asset first.
- `clean_db()` replaces `drop_all`, and the two classes that insert 5000
  contacts refresh the planner statistics afterwards. Both are needed as soon
  as a container outlives one test class, whatever starts it.
- The schema is the one of the migrations, which the db image applies.
  `setUpClass` no longer reflects each database into `Base.metadata` - that was
  there for the removed `drop_all`, and it is what made the metadata describe
  the tables of the other assets - and `test_database.py` no longer rebuilds the
  schema from the models. One DAO test relied on that: the models leave
  `dird_phonebook.tenant_uuid` nullable, the migrations declare it NOT NULL.

**Depends on [wazo-test-helpers#130](https://github.com/wazo-platform/wazo-test-helpers/pull/130)**,
which must be merged first. The plugin used to leave the compose network of
every asset allocated, and 24 assets exhaust the address pool of Docker, so the
**next** run failed. That is fixed upstream now, and this pull request carries
no copy of it. A tox environment that was built before that merge still leaks,
because the pin resolves `master.zip` when the environment is built; recreate it
with `tox -e integration --recreate`. See red bin RB-82.

Result: **24 stack starts**, and the plugin also stops an asset as soon as the
next test needs another one and reports a failed teardown at the end.

## The pattern, and how it differs

A test class in dird names its stack once:

```python
class TestFavorites(BaseDirdIntegrationTest):
    asset = 'all_routes'
```

`__init_subclass__` reads that name and gives the class two things: `asset_cls`,
the class that owns the stack, and the `usefixtures` marker that the plugin reads
to group the classes. A name that no asset knows raises at import.

An asset is a `docker-compose.<name>.override.yml` in `assets/`, and each one now
has a class beside it that names it and nothing else, as wazo-auth and wazo-chatd
do:

```python
class AllRoutesAsset(DirdAssetLaunchingTestCase):
    asset = 'all_routes'
```

`__test__ = False` sits on the base, so the 24 do not repeat it, and
`ASSET_CLASSES` reads the name back from each class rather than list the 24 names
a second time. Two classes that named the same asset would drop one of them from
that map, and its fixture with it, so `test_asset_wiring.py` counts the classes
against the map.

These 24 classes are boilerplate that differ by one string. Merging them is left
for later work.

wazo-auth and wazo-chatd name the stack in three places instead: the asset class
is written by hand, the test base sets `asset_cls`, and each test class carries
`@use_asset('<name>')`. wazo-agid and wazo-sms have one asset each and tests
written for pytest, so they need neither a marker nor an asset class.

| | wazo-dird | wazo-auth, wazo-chatd | wazo-agid, wazo-sms |
| --- | --- | --- | --- |
| assets | 24 | 6-7 | 1 |
| asset classes | 24, written by hand | written by hand | written by hand |
| marker | derived from `asset` | one for each test class | none |
| asset API in the tests | 7 wrappers to `asset_cls` | a wrapper for each method | direct, from the fixture |
| test classes | `unittest.TestCase` | `unittest.TestCase` | functions for pytest |
| plugin | `pytest_asset` | `pytest_asset` | `make_asset_fixture` |

## The decision

dird has more assets than every other project together, and 153 test classes
that already declare the asset they need. To repeat that name in a decorator and
in `asset_cls` would be 153 edits, and it would let the two disagree: the marker
decides which stack starts, `asset_cls` decides which stack the test speaks to.
The 55 marked classes of wazo-auth do agree, but nothing makes them agree. One
declaration cannot disagree with itself.

The wrappers answer a similar question. The test classes *were* the asset class,
so they reach it in 69 places across 21 files, 51 of them for `service_port`.
Seven delegations of one line each keep every one of those calls and give the
asset API a single seam; to write `asset_cls.` at each of the 69 sites would also
tie `helpers/fixtures/http.py` to that attribute, where the decorators only
receive an untyped `self`.

What it costs: the pattern is not the one of the other projects, and it depends
on how pytest collects a marker. pytest takes the markers of the whole MRO and
the plugin keeps the **first**, so only the classes that pytest collects carry
one. A marker on a base class would send every subclass to the stack of that
base class, and the tests would fail far from the cause. 8 classes were found in
that state while this was written. `suite/test_asset_wiring.py` guards both
rules, and it needs no container.

## Notes for the reviewer

- **`mark_logs` is deliberately off.** It marks the logs of `cls.service`, but
  several assets never start a dird container: the backend tests drive the
  backend inside the process of the test. The fixture raises `NoSuchService` on
  each of their tests.
- The two classes of `test_csv_backend.py` that use `asset` only to find a CSV
  file on disk keep no marker, because they need no container.
- `clean_db` uses `DELETE`, not `TRUNCATE`: `TRUNCATE` needs an ACCESS
  EXCLUSIVE lock and would queue behind the queries that dird still has in
  flight. It also deletes only the tables that the database really has, because
  `Base.metadata` belongs to the process and `setUpClass` reflects every
  database into it.

## Tests

Full suite: **570 passed, 0 failed, in 341 s (5 min 41 s)**, which leaves
**0 network** behind and keeps 84 containers to be examined. Measured with the
change of wazo-test-helpers#130 installed, which is what this branch expects.
`suite/test_asset_wiring.py` checks the wiring with no container.

## Follow-ups, out of scope here

1. The models do not describe the schema that we install: they leave
   `dird_phonebook.tenant_uuid` nullable where the migrations declare it NOT
   NULL, and they disagree on the name of the unique
   constraint of `dird_profile` (`wazo_dird/database/models.py:159` against
   `alembic/versions/baseline-2514.sql:289`, compared in
   `wazo_dird/database/queries/profile.py:69` and `:199`). The per-class
   container hid it; a duplicate profile answers 500 in place of 409 once the
   schema comes from the models. Not a production defect, since production uses
   the alembic schema.
2. The compose projects are named `dird_<asset>`, with no checkout in the name,
   so two runs of the suite on one machine share the same containers and
   database and destroy each other. A reused stack lives much longer, which
   widens the window.

---

> [!NOTE]
> **Medium Risk**
> Large test-harness refactor and shared long-lived DB state change how failures propagate and how data is reset; production runtime is unaffected, but incorrect asset wiring or cleanup could cause flaky or cascading integration failures.
> 
> **Overview**
> **Reuses one Docker stack per asset** across integration test classes by adopting `wazo_test_helpers.pytest_asset`, cutting repeated compose up/down from ~153 starts to **24**.
> 
> Asset lifecycle moves to **24 `DirdAssetLaunchingTestCase` subclasses** (one per compose override); `conftest.py` registers session fixtures from `ASSET_CLASSES` and adds an autouse `mark_logs` fixture that skips assets without a `dird` container. Test classes inherit **`DirdAssetRunningTestCase`** (`unittest.TestCase`): declaring `asset = '<name>'` binds `asset_cls` and applies `usefixtures` via `__init_subclass__`, with thin delegators for `service_port`, `docker_exec`, and related asset APIs.
> 
> **Database-backed tests** no longer `reflect`/`create_all`/`drop_all` on teardown; they **`clean_db()`** with row `DELETE` (plus `VACUUM ANALYZE`) on the Alembic schema, and bulk phonebook loads call **`analyze_db()`**. `test_database` / `test_dird_phonebook_backend` replace `setup_module` with a **module-scoped pytest fixture** so the `database` asset starts before the session opens. A **`tenant` test fixture** and a stricter BaseDAO phonebook test align with migration `NOT NULL` on `tenant_uuid`.
> 
> **`test_asset_wiring.py`** asserts each collected `Test*` class’s fixture marker matches its `asset`. OpenAPI doc tests switch from the removed **`documentation`** compose override to **`phonebook_only`**. Minor hardening: `try`/`finally` on several `tearDownClass` paths and `ClassVar` typing where class attributes are set in `setUpClass`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d94ae68095f5100dac584b5401674c1a61aeaa22. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/integration_tests/suite/helpers/base.py
+++ b/integration_tests/suite/helpers/base.py
@@ -11,6 +11,7 @@ from typing import ClassVar
 import requests
 import yaml
 from hamcrest import assert_that, equal_to, has_entries
+from sqlalchemy import inspect, text
 from sqlalchemy.engine import Engine
 from sqlalchemy.orm import scoped_session
 from wazo_dird_client import Client as DirdClient
@@ -75,13 +76,48 @@ class DBRunningTestCase(DirdAssetRunningTestCase):
         database.Base.metadata.create_all(bind=cls.engine)
 
     @classmethod
+    def clean_db(cls) -> None:
+        """Empty every table without touching the schema.
+
+        The schema of the db image comes from the alembic migrations. Dropping
+        it would let `create_all` rebuild a different one from the models - the
+        constraint names do not agree - so the tests would no longer run
+        against the schema that we install.
+        """
+        # `Base.metadata` belongs to the process, and `setUpClass` reflects each
+        # database into it. One session covers many assets, so the metadata
+        # ends up describing tables that the current database does not have.
+        existing = set(inspect(cls.engine).get_table_names())
+
+        with cls.engine.begin() as connection:
+            # DELETE takes a row lock. TRUNCATE needs an ACCESS EXCLUSIVE lock,
+            # so it would queue behind the queries that dird still has in
+            # flight and block the rest of the run.
+            connection.execute(text("SET LOCAL lock_timeout = '10s'"))
+            for table in reversed(database.Base.metadata.sorted_tables):
+                if table.name == 'alembic_version' or table.name not in existing:
+                    continue
+                connection.execute(table.delete())
+
+        # Reclaim the dead rows and refresh the planner statistics. Without
+        # this, a class that loads many contacts leaves the next one with a
+        # bloated table and stale statistics, and the phonebook queries become
+        # slow enough to time out.
+        with cls.engine.connect().execution_options(
+            isolation_level='AUTOCOMMIT'
+        ) as connection:
+            connection.execute(text('VACUUM (ANALYZE)'))
+
+    @classmethod
     def tearDownClass(cls):
+        # A failure of the cleanup must still release the connections and stop
+        # the stack, otherwise the next class of the same asset inherits the
+        # leak and every later class fails too.
         try:
-            database.Base.metadata.drop_all(bind=cls.engine)
-        except Exception:
-            pass
-        cls.engine.dispose()
-        super().tearDownClass()
+            cls.clean_db()
+        finally:
+            cls.engine.dispose()
+            super().tearDownClass()
 
     @classmethod
     def restart_postgres(cls):
--- a/integration_tests/suite/helpers/base.py
+++ b/integration_tests/suite/helpers/base.py
@@ -4,10 +4,12 @@
 import os
 import random
 import string
+import unittest
 from collections.abc import Generator
 from contextlib import contextmanager
-from typing import ClassVar
+from typing import Any, ClassVar
 
+import pytest
 import requests
 import yaml
 from hamcrest import assert_that, equal_to, has_entries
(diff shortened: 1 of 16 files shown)


integration_tests/suite/helpers/base.py
```diff
@@ -74,14 +265,62 @@ def setUpClass(cls):
...
+                connection.execute(table.delete())
+
+        # Reclaim the dead rows and refresh the planner statistics. Without
+        # this, a class that loads many contacts leaves the next one with a
+        # bloated table and stale statistics, and the phonebook queries become
+        # slow enough to time out.
+        with cls.engine.connect().execution_options(
+            isolation_level='AUTOCOMMIT'
+        ) as connection:
+            connection.execute(text('VACUUM (ANALYZE)'))
```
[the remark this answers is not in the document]
DrPyser: Done, using your wording.

integration_tests/suite/helpers/base.py
```diff
@@ -181,6 +377,31 @@ def tearDownClass(cls):
         cls.config.tear_down()
         super().tearDownClass()
```
ja0x:
```suggestion
try:
    cls.config.tear_down()
finally:
    super().tearDownClass()
```
As per the comment at line 298, a failure happening inside `config.tear_down()` will raise and prevent `tearDownClass()` from running (cleandb + engine.dispose).

Didn't dig too deep into `config.tear_down()`, so it might not be able to raise (tbd)

integration_tests/suite/helpers/base.py
ja0x:
```suggestion
try:
    yield
finally:
    cls.start_service('auth')
    auth = cls.make_mock_auth()
    until.true(auth.is_up, timeout=START_TIMEOUT)
    cls.configure_wazo_auth()
```
Same protection to prevent an exception to propagate through the contextmanager and prevent auth from restarting

integration_tests/suite/test_phonebook_http.py
ja0x:
```suggestion
try:
    self.stack.close()
finally:
    super().tearDownClass()
```
Make sure teardownclass runs on failure

ja0x:
```suggestion
try:
    self.stack.close()
finally:
    super().tearDownClass()
```
Make sure teardownclass runs on failure

ja0x:
```suggestion
try:
    self.stack.close()
finally:
    super().tearDownClass()
```
Make sure teardownclass runs on failure

integration_tests/suite/test_phonebook_http.py
```diff
@@ -917,6 +931,7 @@ def setUpClass(cls) -> None:
             ],
         )
         assert not errors
+        cls.analyze_db()
 
     @classmethod
     def tearDownClass(self):
```
ja0x:
```suggestion
    def tearDownClass(cls):
```
typo picked up by Claude

integration_tests/suite/test_phonebook_http.py
ja0x:
```suggestion
    def tearDownClass(cls):
```
typo picked up by Claude

integration_tests/suite/test_database.py
```diff
@@ -80,15 +81,18 @@ class DBStarter(DBRunningTestCase):
...
-def setup_module():
+@pytest.fixture(scope='module', autouse=True)
+def _database_session(request):
+    """Open the session once the `database` stack is up.
+
+    The stack belongs to the asset plugin, so it must be asked for through the
+    fixture; `setup_module` would run before it.
+    """
     global Session
+    request.getfixturevalue('database')
```
ja0x:
```suggestion
```
This method doesn't create the "actual" marker on the class which grouping + teardown are driven by (class markers `_marker_of`). 
Without the mark, the database classes are grouped under an empty '' key, cleaned up at the end by the finally block in the plugin

integration_tests/suite/test_database.py
ja0x:
```suggestion
@pytest.mark.usefixtures('database')
class _BaseTest(unittest.TestCase):
```
The mark must be a class mark, since the pytest_asset plugin reads Class.own_markers and a module level mark is invisible there 

The mark will propagate through the MRO to each database subclasses

ja0x (changes requested):
Very cool, requesting some light harderning recommanded by Claude

integration_tests/suite/test_database.py
DrPyser: yeah was not sure about this. thanks

ja0x (approved):
LGTM, thanks for this optimization

PR #272, opened 2026-08-13
http: fix stop-before-start shutdown race

fblackburn1:
Why: a SIGTERM landing during startup was silently lost when stop()
ran before run() assigned self.server: the stop was a no-op, run()
then started the server with nothing left to stop it, and systemd
escalated to SIGKILL after TimeoutStopSec. The window was the whole
startup phase (plugin loading, source configuration).

stop() now sets a tombstone before stopping the server, and run()
checks it after assigning self.server: whichever side loses the
race, either run() sees the tombstone or stop() sees a real server.

Same fix as wazo-auth (ITEM-589).

---

> [!NOTE]
> **Low Risk**
> Small, localized lifecycle change to HTTP server startup/shutdown with tests; no auth or data-path changes.
> 
> **Overview**
> Fixes a **shutdown race** when `stop()` runs before `run()` has created `self.server` (e.g. SIGTERM during plugin/source startup): the old early `stop()` was a no-op, then `run()` still called `start()`, leaving nothing to shut down cleanly.
> 
> `CoreRestApi` now keeps a **`threading.Event` tombstone** set at the beginning of `stop()`. After building `DynamicWSGIServer` but **before** `start()`, `run()` bails out with a warning if stop was already requested; if `run()` finished setup first, `stop()` still stops the real server as before.
> 
> Adds **unit tests** for stop-before-run, run-after-stop (no `start()`), and stop-after-run (`stop()` on the server).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 922897d68d1fc24e0d67c358c35383e2d2bd1ee5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wazo_dird/http_server.py
+++ b/wazo_dird/http_server.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import logging
 import os
+import threading
 from datetime import timedelta
 
 from flask import Flask
@@ -51,6 +52,7 @@ class CoreRestApi:
         app.config.update(global_config)
         self.load_cors()
         self.server: wsgi.DynamicWSGIServer | None = None
+        self._stopped = threading.Event()
         self.app = app
         self.api = api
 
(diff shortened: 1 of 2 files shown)


ja0x (approved):
LGTM, closes the window down to a non-exploitable theorical timing gap

PR #274, opened 2026-08-17
fix(phonebook): stop duplicating the synthetic id field on contact creation

DrPyser:
> [!NOTE]
> **Medium Risk**
> Ships a data-deleting Alembic migration and changes how phonebook contact fields are written; scope is narrow but upgrades must succeed before the unique index applies.
> 
> **Overview**
> Fixes contact field persistence so each contact gets **one** synthetic `id` row (contact UUID) instead of duplicating it, and updates create/edit to **merge** fields in place rather than replacing the whole collection.
> 
> `Contact.fields` is now a name-keyed collection (`attribute_mapped_collection`), and `_set_contact_fields` adds/updates/removes rows to match the request body while always setting `id` from the contact UUID (ignoring any client `id`).
> 
> **Database:** a migration **dedupes** existing `(name, contact_uuid)` duplicates (keeps lowest `id`), then adds a **unique index** on those columns so duplicates cannot return.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4c73551d06eddd65399b2b16984f37c6919f8d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/alembic/versions/52869e467b80_add_unique_index_contact_fields_name_.py
+++ b/alembic/versions/52869e467b80_add_unique_index_contact_fields_name_.py
@@ -1,0 +1,24 @@
+"""add_unique_index_contact_fields_name_contact_uuid
+
+Revision ID: 52869e467b80
+Revises: dc8580445700
+
+"""
+
+# alembic exposes op as a runtime proxy that mypy cannot see statically
+from alembic import op  # type: ignore[attr-defined]
+
+# revision identifiers, used by Alembic.
+revision = '52869e467b80'
+down_revision = 'dc8580445700'
+
+INDEX_NAME = 'dird_contact_fields__idx__name_contact_uuid'
+TABLE_NAME = 'dird_contact_fields'
+
+
+def upgrade() -> None:
+    op.create_index(INDEX_NAME, TABLE_NAME, ['name', 'contact_uuid'], unique=True)
+
+
+def downgrade() -> None:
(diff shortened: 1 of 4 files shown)


fblackburn1 (changes requested):
A migration script to remove duplicate is missing.
Maybe you plan to add it with a unique constraint in the schema (in another PR?)

alembic/versions/52869e467b80_add_unique_index_contact_fields_name_.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = 'dc8580445700'
+
+INDEX_NAME = 'dird_contact_fields__idx__name_contact_uuid'
+
+
+def upgrade() -> None:
+    # CONCURRENTLY cannot run inside a transaction; dird_contact_fields is a
+    # hot, frequently-written table and a plain CREATE INDEX would hold a
+    # lock for the whole build.
+    with op.get_context().autocommit_block():
```
DrPyser: we assume migrations run during planned maintenance downtime during upgrade.

fblackburn1:
Everything manually tested, migration takes <5s with my 25k users phonebook
:heart:  to avoid to recreate new row every time if already exist

fblackburn1 (approved):

PR #277, opened 2026-08-17
TASK-553 TASK-552 optimize favorites

DrPyser:
- **feat(favorites): add temporary contact_id migration plugin**
- **feat(wazo source): key contacts on the confd user uuid**
- **chore(upgrade): add the favorite contact_id migration script**
- **favorite_migration: use utils.is_uuid**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 * Requests to wazo-auth now default to `localhost:80`, through nginx. Existing
   directory sources pointing at `localhost:9497` are migrated.
 
+* `PUT /0.1/directories/favorites/<source>/<contact_id>` now returns `400` when
+  the source rejects `contact_id`; a `wazo` source rejects an id that names no
+  confd user. Use the `source_entry_id` returned by a lookup for a valid id.
+  Sources that cannot check an id keep accepting any.
+
 * Contacts of a `wazo` source are now identified by the confd user `uuid`
   instead of the confd user `id`. Existing favorites are migrated during the
   upgrade.
@@ -12,9 +17,8 @@
     favorites results, is now the user uuid;
   * `wazoSourceEntryId` in GraphQL results is now the user uuid;
   * `PUT` and `DELETE /0.1/directories/favorites/<source>/<contact_id>` accept
-    the user uuid. A favorite added with the confd user id keeps working for
-    one release, but is no longer listed once that release is out: use the
-    `source_entry_id` returned by a lookup.
+    the user uuid. The confd user id is still accepted for backward compatibility
+    (resolved to the corresponding user uuid).
 
 ## 26.08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
   confd user. Use the `source_entry_id` returned by a lookup for a valid id.
   Sources that cannot check an id keep accepting any.
 
+* `PUT /0.1/directories/favorites/<source>/<contact_id>` now returns `503` when
+  the source cannot be reached, instead of storing a favorite that no listing
+  could resolve. Deleting a favorite of such a source still works.
+
 * Contacts of a `wazo` source are now identified by the confd user `uuid`
   instead of the confd user `id`. Existing favorites are migrated during the
   upgrade.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@
 * Requests to wazo-auth now default to `localhost:80`, through nginx. Existing
   directory sources pointing at `localhost:9497` are migrated.
 
+* Contacts of a `wazo` source are now identified by the confd user `uuid`
+  instead of the confd user `id`. Existing favorites are migrated during the
+  upgrade.
+  * `relations.source_entry_id` of a `wazo` contact, in lookup, reverse and
+    favorites results, is now the user uuid;
+  * `wazoSourceEntryId` in GraphQL results is now the user uuid;
+  * `PUT` and `DELETE /0.1/directories/favorites/<source>/<contact_id>` accept
+    the user uuid. A favorite added with the confd user id keeps working for
+    one release, but is no longer listed once that release is out: use the
+    `source_entry_id` returned by a lookup.
+
 ## 26.08
 
 * New `rest_api.min_threads` option: threads kept ready at all times.
--- a/wazo_dird/plugins/base_plugins.py
+++ b/wazo_dird/plugins/base_plugins.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2024 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2014-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import annotations
--- a/wazo_dird/plugins/base_plugins.py
+++ b/wazo_dird/plugins/base_plugins.py
@@ -144,6 +144,16 @@ class BaseSourcePlugin(metaclass=abc.ABCMeta):
                 results[exten] = entry
         return results
 
+    def canonical_unique_id(self, unique_id: str) -> str | None:
+        """The id this backend stores for `unique_id`, `None` if there is none.
+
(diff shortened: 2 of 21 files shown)


wazo_dird/plugins/wazo_user_backend/plugin.py
```diff
@@ -162,18 +167,67 @@ def match_all(
...
+        try:
+            user = self._client.users.get(unique_id)
+        except RequestException as e:
+            logger.info(
+                '%s: cannot resolve contact id %s to a user uuid, keeping it: %s',
+                self.name,
+                unique_id,
+                e,
+            )
+            return unique_id
```
DrPyser: probably should be treated as error instead and reraised

wazo_dird/plugins/base_plugins.py
```diff
@@ -144,6 +144,16 @@ def match_all(
...
         return results
 
+    def canonical_unique_id(self, unique_id: str) -> str | None:
+        """The id this backend stores for `unique_id`, `None` if there is none.
+
+        The favorites service stores what this returns, so a backend that
+        accepts more than one form of id translates it here and keeps a single
+        form in the database. Backends whose ids have no recognisable shape
+        accept anything, which is what this default does.
+        """
```
DrPyser: Will want to trim that

wazo_dird/plugins/favorite_migration/service.py
```diff
@@ -0,0 +1,218 @@
...
+
+        report['migrated'], report['deduplicated'] = self._dao.migrate(
+            source_uuid, changes
+        )
+        self._dao.drop(source_uuid, drops)
+        return report
+
+    def _fetch_id_to_uuid(self, source: SourceInfo) -> dict[str, str]:
+        client = registry.get(source)
+        users = client.users.list(view='directory', recurse=True)
```
fblackburn1: Hum are you sure clients configured in sources have a valid timeout to use `recurse` query?
i.e. on stack with lot of tenants, this endpoint could be very slow and may timeout easily (which doesn't occur in "normal" use case)
We may want to use a "custom" config timeout for the plugin? 🤷‍♂️

wazo_dird/plugins/wazo_user_backend/plugin.py
```diff
@@ -209,7 +254,7 @@ def _fetch_entries(
 
         return (self._source_result_from_entry(entry, uuid) for entry in entries)
 
-    def _get_uuid(self) -> str:
+    def _get_stack_uuid(self) -> str:
```
fblackburn1: nit: we mostly refer to `stack` to `wazo` (i.e. there is no reference to `stack` in the code)

```suggestion
    def _get_wazo_uuid(self) -> str:
```

wazo_dird/plugins/favorite_migration/service.py
DrPyser: should use pagination indeed

wazo_dird/plugins/favorites_service/plugin.py
```diff
@@ -234,6 +260,15 @@ def remove_favorite(
         if not matching_source:
             raise self.NoSuchSourceException(source_name)
 
+        try:
```
fblackburn1: `remove_favorite` now calls `canonical_unique_id` on every DELETE, which for a `wazo` source is a confd `GET /users/<id>` even when the client already sends the uuid — the round trip returns the same string, and it turns a path that was DB-only on master into one that fails when confd is down: `_confd_user` raises `WazoConfdError`, which is not in the `except (InvalidContactId, SourceUnavailable)` tuple, so the fallthrough never triggers and DELETE returns 503 (the CHANGELOG says deleting still works in that case; the unit test only covers `source_manager.get()` returning `None`).

PUT and DELETE don't need the same thing from the backend: PUT must *validate* (unknown uuid → 400) and translate; DELETE only needs to *translate* the deprecated confd id, since it matches the stored row exactly and a wrong id just deletes nothing.

Suggestion: give the backend a translate-only entry point for delete — a uuid is returned as is with no confd call, a digit id is resolved through confd, anything else is returned as given. That keeps the retrocompat on DELETE for the transition, makes the common uuid delete free again, and once the `isdigit()` transition branch is removed the delete path stops touching confd entirely. The confd round trip that remains (digit ids) should raise `SourceUnavailable` rather than `WazoConfdError` so the existing except here actually catches it — same contract as PUT.

fblackburn1:
`wazo_dird/plugins/default_json/api.yml`, `add_favorite`: the CHANGELOG announces `400` (invalid contact id) and `503` (source unavailable) on this PUT, but the spec still lists only `204`/`409`. `remove_favorite` right below already documents `503` per-operation, so the same here plus a `400` entry would match. `FavoriteID`'s description (`plugins/api/api.yml`) could also say the id to use is the `source_entry_id` returned by a lookup — that's the part a client author has no other way to learn from the spec.

wazo_dird/plugins/wazo_user_backend/plugin.py
```diff
@@ -162,24 +168,63 @@ def match_all(
...
+
+        confd resolves `GET /users/<id_or_uuid>`, so the shape has to be
+        checked here: without it a confd id would stay acceptable after the
+        transition ends.
+        """
+        # --- transition: drop `or unique_id.isdigit()` once every favorite is
+        # migrated; a client still sending a confd id then gets a 400 ---
+        return is_uuid(unique_id) or unique_id.isdigit()
+
+    # NOTE: defined before `list` because the name `list` is shadowed by the
```
fblackburn1: The helper only exists so its `list[SourceResult]` signature is declared before `def list` shadows the name. Inlining the loop into `list()` avoids that: the signature of `list` itself and a local `results: list[SourceResult] = []` in its body both pass mypy, so the method and the NOTE can go.

wazo_dird/plugins/favorite_migration/service.py
fblackburn1: limit 1000

CHANGELOG.md
```diff
@@ -5,6 +5,25 @@
...
+* Contacts of a `wazo` source are now identified by the confd user `uuid`
+  instead of the confd user `id`. Existing favorites are migrated during the
+  upgrade.
+  * `relations.source_entry_id` of a `wazo` contact, in lookup, reverse and
+    favorites results, is now the user uuid;
+  * `wazoSourceEntryId` in GraphQL results is now the user uuid;
+  * `PUT` and `DELETE /0.1/directories/favorites/<source>/<contact_id>` accept
+    the user uuid. The confd user id is still accepted for backward compatibility
+    (resolved to the corresponding user uuid).
+
```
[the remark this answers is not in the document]
fblackburn1: Trimmed to two bullets in acc7dd30.

wazo_dird/plugins/base_plugins.py
fblackburn1: Trimmed in 85b42357.

wazo_dird/plugins/favorite_migration/service.py
fblackburn1: Paginated with limit=1000 in a16925da; it also warns when confd returns fewer users than announced.

wazo_dird/plugins/wazo_user_backend/plugin.py
fblackburn1: Renamed in b5449761.

wazo_dird/plugins/favorites_service/plugin.py
fblackburn1: Done in 7211bc99: DELETE now uses a translate-only `translate_unique_id` — a uuid passes through with no confd call, a digit id is resolved, and a confd failure raises `SourceUnavailable`, caught to delete the id as given. Side effect: the 503 error id is now `source-unavailable` instead of `wazo-confd-error` (confd details go to the log).

wazo_dird/plugins/wazo_user_backend/plugin.py
fblackburn1: Inlined in 367f7a6b.

fblackburn1 (approved):

PR #278, opened 2026-08-18
ITEM-606 ci(zuul): add a job comparing the code models with the migrated schema

DrPyser:
Why: the models could silently describe a schema that differs from the one
the migrations install, as BUG-450 found by chance. This job builds both
schemas in two empty databases and diffs them with migra, so a mismatch
fails the build instead of being found later, if at all.

This job currently fails: fixing what it finds is a follow-up commit.

PRODUCT-336

Depends-On: https://github.com/wazo-platform/wazo-tools/pull/82
Depends-On: https://github.com/TinxHQ/wazo-production-sf-jobs/pull/54

---

> [!NOTE]
> **Medium Risk**
> Adds data-cleaning migrations and new uniqueness/check constraints on profile-related tables, which can delete mismatched rows on upgrade. CI and model naming changes are otherwise low risk.
> 
> **Overview**
> Aligns SQLAlchemy models with the real migrated schema and adds a Zuul `check-db-schema` job so model/migration drift fails CI instead of shipping silently.
> 
> **Models** now name unique/check/PK constraints to match the database, mark several tenant/FK columns `nullable=False`, give `dird_profile_service_source` an explicit composite PK, and bind `dird_phonebook_id_seq` (including `OWNED BY` on `create_all`) so `create_all` matches Alembic.
> 
> **Migrations** add the previously missing `dird_profile` tenant/display check (nulling mismatched display links first) and the `dird_profile_service_source` PK plus tenant-equality check (deduping and dropping cross-tenant rows first).
> 
> Also treats a `None` profile `display` as empty instead of crashing, with unit/integration coverage for that and duplicate profile create.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6190b7d47c927c3818d263cc996cfae06030071b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


DrPyser:
recheck

DrPyser:
recheck

alembic/versions/f9dbff02cc91_add_missing_dird_profile_tenant_check_.py
```diff
@@ -0,0 +1,45 @@
...
+        .select_from(_profile)
+        .where(_profile.c.tenant_uuid != _profile.c.display_tenant_uuid)
+    ).scalar()
+    if violations:
+        raise RuntimeError(
+            f'{violations} row(s) in dird_profile have tenant_uuid != '
+            'display_tenant_uuid; BUG-455 assumed this never happens. '
+            'Investigate before adding this constraint.'
+        )
+
```
fblackburn1: Small preference to remove the row (with a logger)... I guess this row was non-usable anyway? 
The solution for someone in this situation is to delete it (for >90% of the time)
For the remaining 10%, well we have automatic backup enabled

Same for other RuntimeError

I never seen this pattern in our migration scripts, people generally don't want to break their upgrade for dirty data
And Profile information is not high sensitive/unrecoverable data :smile:

zuul.yaml
```diff
@@ -17,3 +19,8 @@
     timeout: 10800
     vars:
       integration_test_timeout: 60
+
+- job:
+    name: dird-check-db-schema
+    description: Compare the schema of the models with the one the migrations install
+    parent: check-db-schema
```
fblackburn1: Use the `check-db-schema-pg15` that has been added in zuul

fblackburn1 (approved):

PR #280, opened 2026-08-25
ITEM-606-review-fixes

fblackburn1:
- **migration: delete NULL-key rows before adding the primary key**
- **fix(database): drop the redundant unique flag on dird_phonebook.id**

---

> [!NOTE]
> **Medium Risk**
> The migration permanently deletes orphan association rows with NULL keys during upgrade; behavior is intentional but affects live data if such rows exist.
> 
> **Overview**
> Fixes two schema/migration issues flagged in review.
> 
> The **`dird_profile_service_source`** Alembic upgrade now **deletes rows with NULL `profile_service_uuid` or `source_uuid`** before deduplication and primary-key creation. Those rows are not enforced by the composite foreign keys (MATCH SIMPLE), and would cause the migration to fail when adding the composite primary key.
> 
> On **`Phonebook`**, the column-level **`unique=True` on `id` is removed**; **`id` uniqueness is unchanged** because it remains defined by the existing `dird_phonebook_id` unique constraint in `__table_args__`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bd54f3cf3c4b251ac64a9b5cebc868d53677c77. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/alembic/versions/30ae0fd93125_add_missing_dird_profile_service_source_.py
+++ b/alembic/versions/30ae0fd93125_add_missing_dird_profile_service_source_.py
@@ -27,6 +27,18 @@ _profile_service_source = sa.table(
 def upgrade() -> None:
     connection = op.get_bind()
 
+    # The composite FKs use MATCH SIMPLE, so a NULL in either key column
+    # escapes FK enforcement; such orphans would make the primary key
+    # creation below abort.
+    connection.execute(
+        _profile_service_source.delete().where(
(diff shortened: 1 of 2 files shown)


DrPyser (approved):
LGTM, good catch.

PR #281, opened 2026-08-27
BUG-465: reject limit=0 on phonebook contact listing

DrPyser:
## Why

`limit=0` meant two different things depending on which contact list you asked for:

- `GET /0.1/phonebooks/{uuid}/contacts?limit=0` → **200 with every contact**. The phonebook path only applied the limit when truthy, so `limit=0` applied none — returning the large response ITEM-582 set out to avoid.
- `GET /0.1/personal?limit=0` → **400** `"limit must be a positive number"`, from `BaseDAO.validate_parameters`.

Reported as **BUG-465**.

## What changed

`limit=0` is now rejected on both, and both express the rule in a marshmallow schema rather than one doing it at the HTTP layer and the other in the DAO.

**Phonebook** — `limit` overridden with `Range(min=1)` on both contact-listing schemas:
- `phonebook/schemas.py::ContactListSchema` — `GET /phonebooks/{uuid}/contacts`
- `phonebook_backend/schemas.py::ContactListSchema` — `GET /backends/phonebook/sources/{uuid}/contacts` (same DAO, different route)

The override uses `xivo.mallow`'s `fields.Integer`/`validate.Range` rather than plain marshmallow, matching the upstream `ListSchema` definition so the 400 body keeps wazo's structured `constraint_id`/`constraint` error format.

**Personal** — new `personal/schemas.py` validating `limit` (`Range(min=1)`) and `offset` (`Range(min=0)`), applied in `PersonalAll.get`. Two constraints shaped this:
- The personal resource is a `LegacyAuthResource`, whose base has no `handle_validation_exception`, so a `ValidationError` would otherwise surface as a 500. It is caught locally and returned in personal's existing `{reason, timestamp, status_code}` shape — response format is unchanged.
- `order` deliberately stays in the DAO. Personal contact fields are free-form (arbitrary imported columns), so there is no column list to validate against in a schema; validation there is necessarily data-dependent.

Not changed: the **deprecated** phonebook API validates through its own `_ArgParser` instead of a marshmallow schema, and still drops a falsy limit. Left alone — different validation path, and deprecated.

## Behaviour change

`limit=0` on the phonebook contact-listing endpoints goes from `200` + all contacts to `400`. Callers sending `limit=0` and relying on getting everything back must omit `limit` instead. Personal's outcome is unchanged (already 400).

## Note on the DAO layer

Now that #273 is merged, `PhonebookContactCRUD._paginate_contact_uuids` still treats a falsy limit as "no limit" (`if limit:`), and `test_that_a_zero_limit_means_no_limit` asserts that at the DAO level. Both remain valid — the DAO can no longer receive `limit=0` through the API, so the two layers do not conflict — but the DAO permissiveness is now unreachable from HTTP and carries a `TODO` marking it for follow-up.

## Validation

- Red/green verified: with the phonebook fix reverted its test fails (`Expected: <400>`, got 200) while personal already passes — isolating exactly the reported inconsistency.
- Linters, black, **mypy**, isort — green.
- Integration: **142 passed** across `test_phonebook.py`, `test_personal.py`, `test_phonebook_http.py`, `test_dird_phonebook_backend.py`, `test_phonebook_deprecated.py` — re-run after rebasing onto the post-#273-merge base. The existing test pinning personal's `order` error body still passes, confirming the response format is untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/integration_tests/suite/test_personal.py
+++ b/integration_tests/suite/test_personal.py
@@ -45,6 +45,11 @@ class TestListPersonal(PersonalOnlyTestCase):
 
         assert_that(result['items'], contains_exactly())
 
+    def test_that_a_zero_limit_returns_400(self):
+        result = self.list_personal_result(VALID_TOKEN_MAIN_TENANT, limit=0)
+
+        assert_that(result.status_code, equal_to(400))
+
 
 class TestDeletedUser(BaseDirdIntegrationTest):
     asset = 'personal_only'
--- a/integration_tests/suite/test_personal.py
+++ b/integration_tests/suite/test_personal.py
@@ -45,10 +45,18 @@ class TestListPersonal(PersonalOnlyTestCase):
 
         assert_that(result['items'], contains_exactly())
 
-    def test_that_a_zero_limit_returns_400(self):
-        result = self.list_personal_result(VALID_TOKEN_MAIN_TENANT, limit=0)
+    def test_that_an_invalid_limit_or_offset_returns_400(self):
+        for value in [0, -1]:
+            result = self.list_personal_result(VALID_TOKEN_MAIN_TENANT, limit=value)
+            assert_that(result.status_code, equal_to(400), f'limit={value}')
 
+        result = self.list_personal_result(VALID_TOKEN_MAIN_TENANT, offset=-1)
         assert_that(result.status_code, equal_to(400))
+
+    def test_that_a_valid_limit_and_offset_are_accepted(self):
+        result = self.list_personal_result(VALID_TOKEN_MAIN_TENANT, limit=1, offset=0)
+
+        assert_that(result.status_code, equal_to(200))
 
 
 class TestDeletedUser(BaseDirdIntegrationTest):
--- a/integration_tests/suite/test_phonebook.py
+++ b/integration_tests/suite/test_phonebook.py
@@ -452,6 +452,15 @@ class TestContactGet(_BasePhonebookContactTestCase):
 
 
 class TestContactList(_BasePhonebookContactTestCase):
+    def test_that_a_zero_limit_returns_400(self):
+        self.set_tenants(self.tenant_1.name)
+
+        result = self.list_phonebook_contacts(
+            self.phonebook_1['uuid'], tenant=self.tenant_1.uuid, limit=0
+        )
+
+        assert_that(result.status_code, equal_to(400))
+
     def test_unknown_tenant_or_phonebook(self):
         self.set_tenants(self.tenant_2.name)
         result = self.list_phonebook_contacts(
--- a/wazo_dird/plugins/personal/http.py
+++ b/wazo_dird/plugins/personal/http.py
@@ -13,10 +13,13 @@ from typing import TYPE_CHECKING, Any
 
 from flask import Response, request
 from flask_restful import reqparse
+from marshmallow import ValidationError
 from xivo.tenant_flask_helpers import Tenant
 
 from wazo_dird.auth import get_user_uuid, required_acl
 from wazo_dird.http import LegacyAuthResource, get_json_body
+
+from .schemas import list_schema
 
 if TYPE_CHECKING:
     from wazo_dird.database.queries.base import ContactInfo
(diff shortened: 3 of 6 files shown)


fblackburn1:
Can you make a quick pass on all dird endpoints to have same behavior with `limit`

DrPyser:
> Can you make a quick pass on all dird endpoints to have same behavior with `limit`

looking at this further, seems more sensible to actually implement `limit=0` as `LIMIT 0` in sql instead of rejecting with validation error.
This is in line with the majority precedent in wazo-dird and other services. 
xivo-lib-python schema defines `limit = Range(min=0)`, so dird is going against the grain by rejecting `0` as an invalid value.

fblackburn1:
> > Can you make a quick pass on all dird endpoints to have same behavior with `limit`
> 
> looking at this further, seems more sensible to actually implement `limit=0` as `LIMIT 0` in sql instead of rejecting with validation error. This is in line with the majority precedent in wazo-dird and other services. xivo-lib-python schema defines `limit = Range(min=0)`, so dird is going against the grain by rejecting `0` as an invalid value.

Well it's really not consistent across all wazo services ... :persevere: 
For the use case of `limite=0`, I already see it when building a dashboard where you only want the total number. I don't know if it's the right way, but it's convenient :sweat_smile:

DrPyser:
@fblackburn1 I made the pass on all dird endpoints. The pass shows that `limit=0` has four different meanings today. Here is the plan to make one behaviour.

## Proposed rule

- `limit` is an integer that is 0 or more.
- `limit=0` gives an empty page.
- No `limit` parameter gives no limit.
- `offset` is an integer that is 0 or more.

`xivo.mallow_helpers.ListSchema` declares `Range(min=0)` for `limit`. This rule therefore keeps dird the same as the other wazo services, and the same as SQL `LIMIT 0`.

This rule replaces the rule in this PR. This PR rejects `limit=0` with HTTP 400 on three endpoints, which makes dird different from the platform.

In the code, the rule is simple: a limit condition must test `is not None`. It must not test the truth of the value. Each location that is different has a condition `if limit:`, which ignores the value 0.

## Endpoints that are already correct

These endpoints give an empty page for `limit=0`. They need no change. I will add a test for each one.

- `GET /phonebooks`
- `GET /displays`
- `GET /profiles`
- `GET /sources`
- `GET /backends`
- `GET /backends/{name}/sources`
- `GET /directories/{profile}/sources`
- `GET /backends/google/sources/{uuid}/contacts`
- `GET /backends/conference/sources/{uuid}/contacts` and `GET /backends/wazo/sources/{uuid}/contacts`. These two endpoints send the request to wazo-confd, which applies the same schema.

## Locations that are different

| Location | Behaviour today | Change |
| --- | --- | --- |
| `database/queries/phonebook.py:483` | `limit=0` gives all contacts | `if limit is not None:` |
| `database/queries/phonebook.py:166` | `if limit:` | `if limit is not None:` |
| `database/queries/personal.py:172` | the slice `offset + limit if limit else None` gives all contacts | test `if limit is not None` |
| `database/queries/personal.py:100` | `if limit:` | `if limit is not None:` |
| `database/queries/base.py:180` | `limit <= 0` gives HTTP 400 | `limit < 0` |
| `plugins/office365_backend/http.py:55` | `_map_list_params` removes the value 0 before the schema. The default `None` then gives no limit. | `if limit is not None:` |
| `plugins/phonebook_deprecated/http.py:92` | `if self._limit:` removes the value 0, which gives all contacts | `if self._limit is not None:` |

## Changes to remove from this PR

The two commits in this PR apply the opposite rule. I will remove these four changes:

- the `limit` override in `plugins/phonebook/schemas.py`
- the `limit` override in `plugins/phonebook_backend/schemas.py`
- the new file `plugins/personal/schemas.py`
- the schema code in `plugins/personal/http.py`

The branch is not merged. I will therefore rebase on master and write the commits again.

## Steps

1. Rebase the branch on master.
2. Write the tests first. Note that `test_database.py::test_that_a_zero_limit_means_no_limit` asserts the incorrect behaviour. I must change this test to assert an empty result.
3. Change the seven locations in the table.
4. Remove the `TODO` comment at `database/queries/phonebook.py:482`. This change makes the comment obsolete.
5. Run the linters and the integration tests.

## Effect on the API

- `GET /phonebooks/{uuid}/contacts?limit=0` gives an empty page. Before, it gives all contacts.
- The deprecated phonebook endpoints do the same.
- `GET /personal?limit=0` gives HTTP 200 and an empty page. Before, it gives HTTP 400.

A client that sends `limit=0` to get all contacts must remove the `limit` parameter.

This is a change to the API. I will therefore add an entry to `CHANGELOG.md`.

Tell me if you agree with the rule. I will then do the work on this branch.

fblackburn1:
Proposition LGTM

> No limit parameter gives no limit.

Until we add hard limit behavior (ex: returning 25k phonebook contacts may be outside of an *acceptable* range, so hard limit should be followed). Same as external phonebook API provider

